PR #79, opened 2024-04-20
Several fix

jinningwang:
- Fix OModel parse status to ensure no_parsed params can be updated
- Fix and rerun ex2
- Format ``Routine.get()`` return type to be consistent with input idx type
- Remove unused ``Routine.prepare()``


jinningwang (commented):
Good to go.

jinningwang (commented):
Good to go.

PR #244, opened 2026-05-02
refactor(opt): cvxpy-namespace passthrough — routine migration + rewriter cleanup

leocuisine:
## Summary

Step 1 + Step 2 of the cvxpy-namespace passthrough plan, bundled into a single squashed commit.

- **Routines** (13 files, ~73 strings): `mul(...)` → `cp.multiply(...)`; bare `multiply(`/`sum(` → `cp.multiply(`/`cp.sum(`; `<a> dot <b>` → `<a> * <b>`.
- **Generator** (`ams/prep/generator.py`): drop `_FUNCTION_REWRITES`, `_RESERVED_SYMBOL_NAMES`, the symbol-name collision check, and the function-rewrite loop in `_rewrite()`. `_build_symbol_regex` now emits `(?<!\.)\b...\b` for symbol resolution only. Add `PYCODE_FORMAT_VERSION = 2` to the generated header.
- **SymProcessor** (`ams/core/symprocessor.py`): drop the static function-rewrite block from `sub_map`; add `(?<!\.)` lookbehind to every per-symbol pattern.
- **Cache invalidation** (`RoutineBase` + `ams.prep.prep_all`): reject caches whose `pycode_format_version` doesn't match current — auto-heal, mirrors the pristine marker.

Net diff: 16 files, +190 / -191.

## Test plan

- [ ] CI green (pytest + flake8 + Codacy)
- [ ] `pytest tests/` passes locally (346 tests previously)
- [ ] `flake8 ams/` clean
- [ ] pycode parity vs pre-Step-1 baseline: 33/33 byte-identical (excluding `timestamp` / `md5` / `ams_version` / `pycode_format_version`)
- [ ] Stale-cache auto-rebuild verified: corrupting `pycode_format_version` triggers regen on `ams prep` (no `--force`)


leocuisine (commented):
<h2>Independent review of PR #244</h2>

The diff is conceptually clean: routines now author canonical CVXPY directly (cp.sum, cp.multiply, *), the old function-name rewrite stage is gone from both SymProcessor.sub_map and generator._FUNCTION_REWRITES, and a new pycode_format_version = 2 header field auto-invalidates pre-Step-2 caches. The mechanical migration of 13 routine files looks correct. What is missing is the user-facing migration: tests, docs, and examples still use the old mul / sum / dot vocabulary, and there is no automated coverage for the new cache-invalidation key.

<h3>Findings</h3>

1. **[bug] tests/test_routine_ns.py:175,207** — Both test_e_str_append_then_reinit and test_customization_does_not_pollute_other_instances append " + 0 * sum(pg)" or " + 1e-6 * sum(pg)" to obj.e_str, then re-init and solve. Before this PR, sub_map contained the rewrite r"\bsum\b" -> "cp.sum", so the eval-fallback path produced cp.sum(self.om.pg). After this PR, that rewrite is gone — the appended sum(self.om.pg) reaches eval and resolves to Python builtin sum, which iterates the CVXPY Variable. Best case it accidentally works (1-D var, builtin sum returns a CVXPY expression by elementwise +); more likely the resulting problem is no longer DPP, fails to solve, or solves to a different value than the assertion expects. Migrate these to cp.sum(pg) to match the new contract, and consider asserting that "cp.sum" appears in obj.code after parse so any future regression is unambiguous.

2. **[api-break] docs/source/modeling/routine.rst:31,104-127,187-188** — The "Authoring e_str" section still documents mul(a, b), multiply(a, b), dot, sum(x), norm(x), pos(x), square(x), power(x, n), etc. as supported syntax, with line 31 carrying a worked example "gs @ mul(ug, pru) - dud". With this PR none of those tokens are rewritten anymore — a user who copies that example into addConstrs(e_str=...) gets a NameError: name "mul" is not defined at eval time. This is the documented public extension API; the migration needs a corresponding rewrite of this page, or at minimum a Migration callout above the existing table noting that the user must now write cp.multiply / cp.sum directly.

3. **[api-break] docs/source/modeling/example.rst:120-122** — Worked example builds an objective as "sum(mul(c2, power(pg, 2)))" + "sum(mul(c1, pg))" + "sum(mul(ug, c0))". Same break as above — no longer parses. Should be updated to the canonical cp.sum(cp.multiply(...)) form, otherwise the very first example a new user runs from the docs raises.

4. **[test-gap] tests/codegen/test_generator.py** — test_header_metadata checks md5, ams_version, cvxpy_version but not the newly-added pycode_format_version. More importantly, there is no test exercising the staleness check itself: write a cache file with a stale pycode_format_version, call _link_pycode / prep_all, assert the file is regenerated. The PR description acknowledges this verification is currently manual — given that auto-invalidation is the entire point of bumping the version, an automated test would catch a future regression where the staleness check is silently skipped.

5. **[correctness-risk] ams/prep/generator.py:114 + ams/core/symprocessor.py:138-145** — Removing _RESERVED_SYMBOL_NAMES is justified for the codegen path because the (?<!\.)\b<name>\b lookbehind skips attribute access. But the eval-fallback path (opt/constraint.py, opt/objective.py, opt/expression.py, opt/exprcalc.py) is still vulnerable: a routine that declares a symbol named, say, sum will now have sub_map[r"(?<!\.)\bsum\b"] = "self.om.sum". A user who appends " + cp.sum(pg)" to obj.e_str is fine because the lookbehind protects cp.sum, but a user who writes the historically-documented " + sum(pg)" gets self.om.sum(self.om.pg) — self.om.sum is the user symbol, not Python builtin, so this silently produces nonsense rather than raising. Worth either keeping a reduced collision check that rejects symbol names matching common CVXPY atoms (sum, multiply, vstack, ...), or at least documenting this trap in the release notes alongside the migration guidance.

6. **[docs] docs/source/release-notes.rst:12** — The release-notes entry was demoted from v1.2.2 (2026-05-01) to v1.2.2 (unreleased), which is fine, but the substantive change in this PR (the e_str authoring contract for downstream users — see findings 2/3/5) is not mentioned. A user upgrading minor versions would not learn from the changelog that their addConstrs(e_str="mul(...)") snippets stop working.

<h3>Non-blocking observations</h3>

- **examples/ex8.ipynb cell 9** — The edit converted the source field from a list-of-lines (the format every other cell in the notebook uses, and what Jupyter writes when it saves) to a single string with embedded \n. The notebook also lost its trailing newline ("No newline at end of file" in the diff). Both are valid nbformat but will produce a noisy diff the next time anyone opens and saves the notebook. Re-saving the notebook through Jupyter once would normalize it.
- **ams/prep/__init__.py:115** — The "from ams.prep.generator import PYCODE_FORMAT_VERSION" line is inside the per-routine loop. The import is cached so the cost is negligible, but hoisting it to the function top would match how the other helpers (source_md5, pycode_dir, _cp) are imported in this file.
- **ams/prep/generator.py:62-64** — The two-rule multiply(...) / \bmultiply\b(...) chain is documented as a workaround for OrderedDict key dedup. The same workaround for mul / \bmul\b was already present and is pre-existing, but the underlying limitation (a single-pass regex cannot flatten arbitrary nesting; the second pass only catches one extra level) means deeply nested cp.multiply(cp.multiply(cp.multiply(a, b), c), d) still will not render fully in tex. Not a regression — just inherited.
- **ams/core/symprocessor.py:51-60** — The "sub_map = OrderedDict()" initializer plus a long explanatory comment plus the empty-then-populated-in-generate_symbols pattern is more cognitive load than necessary. Could be deleted entirely (left implicit) or condensed; the current shape reads as if the comment is doing more work than the code.

<h3>Looks good</h3>

- The (?<!\.) lookbehind is exactly the right hammer for "preserve attribute access during symbol resolution" — it is applied uniformly in both SymProcessor.sub_map (via generate_symbols) and generator._build_symbol_regex, so the codegen and eval-fallback paths stay aligned.
- The pycode_format_version field plus the getattr(gen, "pycode_format_version", None) != PYCODE_FORMAT_VERSION check in _link_pycode is the correct auto-heal pattern, mirrors the existing pristine marker, and treats absent-attribute caches as stale (not as a crash).
- The mechanical migration of all 13 routine files looks consistent — mul to cp.multiply, bare sum to cp.sum, dot to *, with no leftover legacy tokens in ams/routines/.

leocuisine:
Thanks for the review — all 6 findings addressed in `09c70d70`. Per-finding response:

1. **[bug] tests/test_routine_ns.py** — Fixed: `sum(...)` / `mul(...)` in the three customization strings (lines 118, 175, 207) now use `cp.sum` / `cp.multiply`. Did not add the suggested `cp.sum in obj.code` regression assertion: `Objective.parse()` returns early without setting `self.code` when `e_fn is None` is the dirty path here, so the signal isn't reliably available. The existing `exit_code == 0` plus the downstream `test_customization_does_not_pollute_other_instances` (which now also uses canonical CVXPY) cover the behavior.

2. **[api-break] docs/source/modeling/routine.rst** — Rewrote the "Authoring e_str" section to canonical CVXPY (`cp.multiply`, `cp.sum`, etc.). Added a `.. note::` migration callout and a `.. warning::` listing the reserved CVXPY-atom names. Updated the data-section example (`gs @ cp.multiply(ug, pru) - dud`) and the sub_map description.

3. **[api-break] docs/source/modeling/example.rst** — Worked example objective updated to `cp.sum(cp.multiply(...))`.

4. **[test-gap] tests/codegen/test_generator.py** — Added `pycode_format_version` field check to `test_header_metadata`, plus a new `TestStaleCacheRegen` class with two tests (stale field value, missing field). The tests required defeating Python's `__pycache__/.pyc` shadow cache between in-test rewrites — see the helper `_force_source_reload`. The production `_link_pycode` staleness check is correct.

5. **[correctness-risk] generator.py + symprocessor.py] — Re-introduced a reduced collision check via `RESERVED_CVXPY_ATOM_NAMES` and `_check_reserved_collisions(routine, names)` in `ams/prep/generator.py`. Both the codegen path (`_collect_symbol_names`) and the eval-fallback path (`SymProcessor.generate_symbols`) consult it; a routine that declares a symbol named `sum`/`multiply`/`vstack`/etc. now raises an explanatory `ValueError` at codegen / sub_map build time. Confirmed no existing routine collides via an AST sweep.

6. **[docs] release-notes.rst** — Added a "Migration — `e_str` authoring contract" section above the v1.2.2 New features list. Includes the substitution table (`mul → cp.multiply`, `sum → cp.sum`, `dot → *`) and a note about the reserved-name guard.

**Non-blocking observations:**

- *ex8.ipynb cell-9 source-list-vs-string format* — Acknowledged. The `NotebookEdit` tool produced a single-string `source` field; will normalize on the next interactive notebook open. Not changed in this push.
- *`PYCODE_FORMAT_VERSION` import inside the loop in `ams/prep/__init__.py:115`* — Acknowledged but not changed; the import is module-cached and the call site is per-routine, not per-iteration in a hot loop. Will pick up if the surrounding code is touched.
- *Two-rule `multiply(...)` chain in the tex pipeline* — Pre-existing limitation acknowledged; not in scope here.
- *`sub_map = OrderedDict()` initializer style nit in symprocessor* — Left as is.

PR #245, opened 2026-05-02
refactor(opt): collapse legacy e_str eval into eval_e_str helper

leocuisine:
## Summary

- Step 3 of the **cvxpy-namespace passthrough** project (follow-up to PR #244).
- New `ams/opt/_runtime_eval.py::eval_e_str(item, e_str)` helper does symbol resolution (via `_build_symbol_regex` / `_collect_symbol_names`) + `eval` against `{cp, np, sps, r=RoutineNS, self}` with one `# nosec B307`.
- Four near-duplicate `regex + eval` blocks in `opt/{constraint,objective,expression,exprcalc}.py` collapse into one helper call each. Net **−43 lines** across the four files (+71 in the helper).
- `parse()` in each item now stores `self.code = self.e_str` verbatim — symbol rewriting is deferred to evaluate time. `OptzBase.e` and `Routine.formulation_summary` keep working.

`SymProcessor.sub_map` / `val_map` are still populated; **Step 4 retires them** (next PR will add `eval_e_str_numeric` for `OptzBase.e` and delete `sub_map`). The `formulation_source` bucket still reports `'sub_map'` for eval-fallback items — Step 5 renames it to `'eval'`.

## Why this is safe

The new helper resolves `r.<name>` against `RoutineNS`, which walks the same six-tier symbol order (`vars` → `rparams` → `services` → `exprs` → `constrs` → `config`) the legacy `sub_map` regex did. The reserved-name guard from PR #244 (`RESERVED_CVXPY_ATOM_NAMES`) already prevents a routine symbol from shadowing a CVXPY atom. `_build_symbol_regex` carries the `(?<!\.)` lookbehind so canonical `cp.X(...)` survives.

The Constraint LHS-vs-Constraint dispatch (`isinstance(result, cp.constraints.Constraint)`) is now shared between the `e_fn` and `e_str` paths, removing the previous inconsistency where the two branches had subtly different wrapping logic.

## Test plan

- [x] `tests/test_routine_ns.py` (17/17) — covers the post-init `obj.e_str = ...` + re-init customization pattern; this is the only suite that exercises the eval-fallback path on a real routine
- [x] Full `pytest tests/` — 348 passed (no regressions)
- [x] `flake8 ams/opt/` — clean
- [x] Live sanity: `DCOPF.obj.e_str = ... + ' + 0 * cp.sum(pg)'` → re-init → solve. `formulation_source == 'sub_map'`, `exit_code == 0`, objective matches the codegen path bit-for-bit.

### DIFF
--- a/ams/opt/_runtime_eval.py
+++ b/ams/opt/_runtime_eval.py
@@ -1,5 +1,5 @@
 """
-Runtime ``e_str`` evaluation helper.
+Runtime ``e_str`` evaluation helpers.
 
 Used by :class:`Constraint`, :class:`Objective`, :class:`Expression`,
 and :class:`ExpressionCalc` whenever the author/user supplies an
@@ -8,11 +8,15 @@ and :class:`ExpressionCalc` whenever the author/user supplies an
 that previously lived in those classes — see step 3 of the
 cvxpy-namespace passthrough project.
 
-The helper does **only** symbol resolution
-(``\\b<name>\\b → r.<name>``), not function-name rewriting: callers
-are expected to author canonical CVXPY syntax (``cp.sum(...)``,
-``cp.multiply(...)``) since the function-rewrite layer was deleted
-in PR #244.
+Two helpers, both symbol-resolution-only (no function-name rewrites
+— callers must author canonical CVXPY syntax since the
+function-rewrite layer was deleted in PR #244):
+
+- :func:`eval_e_str` — CVXPY-side; produces ``cp.Expression`` /
+  ``cp.Constraint`` for the optimization model.
+- :func:`eval_e_str_numeric` — numeric-side; used by
+  :pyattr:`OptzBase.e` to recover the post-solve numpy LHS for
+  customized items.
 """
 
 import logging
@@ -21,14 +25,38 @@ import cvxpy as cp
 import numpy as np
 
 from ams.shared import sps
-from ams.core.routine_ns import RoutineNS
+from ams.core.routine_ns import RoutineNS, NumericRoutineNS
 from ams.prep.generator import _build_symbol_regex, _collect_symbol_names
 
 logger = logging.getLogger(__name__)
 
 
+def _get_symbol_regex(rtn):
+    """Memoized symbol-resolution regex for ``rtn``.
+
+    Cached on ``rtn.syms`` so the per-evaluate cost of
+    :func:`_collect_symbol_names` (sort + reserved-name set
+    intersection) is paid once per symbol-generation pass, not per
+    item. ``SymProcessor.generate_symbols`` clears the cache when
+    symbols are regenerated.
+    """
+    syms = rtn.syms
+    cached = getattr(syms, '_eval_symbol_regex', None)
+    if cached is not None:
+        return cached
+    regex = _build_symbol_regex(_collect_symbol_names(rtn))
+    syms._eval_symbol_regex = regex
+    return regex
+
+
+def _resolve(rtn, e_str):
+    """Apply the symbol-resolution regex to ``e_str``."""
+    sym_re = _get_symbol_regex(rtn)
+    return sym_re.sub(r'r.\1', e_str) if sym_re is not None else e_str
+
+
 def eval_e_str(item, e_str):
-    """Resolve symbols in ``e_str`` and ``eval`` the result.
+    """Resolve symbols in ``e_str`` and ``eval`` against CVXPY namespace.
 
     Parameters
     ----------
@@ -54,8 +82,7 @@ def eval_e_str(item, e_str):
         isn't already a ``cp.constraints.Constraint``).
     """
     rtn = item.om.rtn
-    sym_re = _build_symbol_regex(_collect_symbol_names(rtn))
-    code = sym_re.sub(r'r.\1', e_str) if sym_re is not None else e_str
+    code = _resolve(rtn, e_str)
     local_vars = {
         'cp': cp, 'np': np, 'sps': sps,
         'r': RoutineNS(rtn), 'self': item,
@@ -73,3 +100,43 @@ def eval_e_str(item, e_str):
             f"  rewritten: {code}\n"
             f"  error: {exc}"
         ) from exc
+
+
+def eval_e_str_numeric(item, e_str):
+    """Evaluate ``e_str`` numerically for :pyattr:`OptzBase.e`.
+
+    Same shape as :func:`eval_e_str`, but resolves symbols against a
+    :class:`NumericRoutineNS` that returns numpy values for Vars,
+    RParams, services, and recomputes Expressions through their
+    ``e_fn``. CVXPY atoms (``cp.sum``, ``cp.multiply``, …) wrap
+    numpy inputs in ``cp.Constant``, so the returned object's
+    ``.value`` is the numpy LHS — equivalent to what the legacy
+    ``val_map`` + ``np.<atom>`` rewrite produced, but driven by the
+    same name-resolution surface as the CVXPY-side helper.
+
+    For a ``cp.constraints.Constraint`` result (e.g. an ``e_str``
+    that already contains ``<= 0``), returns the constraint's LHS
+    expression so callers can ``.value`` it uniformly.
+    """
+    rtn = item.om.rtn
+    code = _resolve(rtn, e_str)
+    local_vars = {
+        'cp': cp, 'np': np, 'sps': sps,
+        'r': NumericRoutineNS(rtn), 'self': item,
+    }
+    try:
+        result = eval(code, {}, local_vars)  # nosec B307
+    except Exception as exc:
+        raise Exception(
+            f"Error numerically evaluating {item.class_name} "
+            f"<{item.name}> via e_str.\n"
+            f"  e_str: {e_str}\n"
+            f"  rewritten: {code}\n"
+            f"  error: {exc}"
+        ) from exc
+    if isinstance(result, cp.constraints.Constraint):
+        inner = getattr(result, '_expr', None)
+        if inner is None:
+            inner = result.args[0] if result.args else None
+        return inner
+    return result
--- a/ams/opt/_runtime_eval.py
+++ b/ams/opt/_runtime_eval.py
@@ -1,0 +1,75 @@
+"""
+Runtime ``e_str`` evaluation helper.
+
+Used by :class:`Constraint`, :class:`Objective`, :class:`Expression`,
+and :class:`ExpressionCalc` whenever the author/user supplies an
+``e_str`` string instead of (or to override) the codegen-wired
+``e_fn``. Replaces the four near-duplicate ``regex + eval`` blocks
+that previously lived in those classes — see step 3 of the
+cvxpy-namespace passthrough project.
+
+The helper does **only** symbol resolution
+(``\\b<name>\\b → r.<name>``), not function-name rewriting: callers
+are expected to author canonical CVXPY syntax (``cp.sum(...)``,
+``cp.multiply(...)``) since the function-rewrite layer was deleted
+in PR #244.
+"""
+
+import logging
+
+import cvxpy as cp
+import numpy as np
+
+from ams.shared import sps
+from ams.core.routine_ns import RoutineNS
+from ams.prep.generator import _build_symbol_regex, _collect_symbol_names
+
+logger = logging.getLogger(__name__)
+
+
+def eval_e_str(item, e_str):
+    """Resolve symbols in ``e_str`` and ``eval`` the result.
+
+    Parameters
+    ----------
+    item : ams.opt.OptzBase
+        The opt-layer item (Constraint/Objective/Expression/
+        ExpressionCalc) whose ``e_str`` is being evaluated. Used both
+        to locate the owner routine (``item.om.rtn``) and to expose
+        ``self`` in the eval namespace for back-compat with any
+        author customization that reaches for it.
+    e_str : str
+        The expression string. Must be in canonical CVXPY syntax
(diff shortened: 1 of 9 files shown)


ams/opt/constraint.py
```diff
@@ -91,21 +91,18 @@ def get_all_idxes(self):
...
-        for pattern, replacement in sub_map.items():
-            try:
-                code_constr = re.sub(pattern, replacement, code_constr)
-            except TypeError as e:
-                raise TypeError(f"Error in parsing constr <{self.name}>.\n{e}")
-        # parse the constraint type
-        code_constr += " == 0" if self.is_eq else " <= 0"
-        # store the parsed expression str code
-        self.code = code_constr
+        self.code = self.e_str
```
leocuisine, line 105: **Finding #1 anchor** — `correctness-risk`. After this line `self.code` is the unrewritten `e_str` (no `self.om.<name>` references and no trailing `== 0` / `<= 0`). `OptzBase.e` (`ams/opt/optzbase.py:315`) still runs `val_map` over `self.code` and `eval`s it with locals `{self, np, cp, val_map}` — no `rtn`. The old shape worked because `sub_map` had already pre-rewritten symbols to `self.om.pg`, which `val_map` then turned into `self.om.rtn.pg.v` (resolvable via `self`). Now `pg + 1` becomes `rtn.pg.v + 1` and `eval` raises `NameError: name 'rtn' is not defined`. Worse, `no_parse` rparams (`c2`) aren't in `val_map` at all, so they stay bare and produce `NameError: name 'c2' is not defined`. `OptzBase.e` swallows the exception and silently returns `None` for any item that goes through the eval-fallback path. See top-level review body for repro + suggested fixes.

ams/opt/_runtime_eval.py
```diff
@@ -0,0 +1,75 @@
...
+        Objective, ``== 0`` / ``<= 0`` for Constraint when the result
+        isn't already a ``cp.constraints.Constraint``).
+    """
+    rtn = item.om.rtn
+    sym_re = _build_symbol_regex(_collect_symbol_names(rtn))
+    code = sym_re.sub(r'r.\1', e_str) if sym_re is not None else e_str
+    local_vars = {
+        'cp': cp, 'np': np, 'sps': sps,
+        'r': RoutineNS(rtn), 'self': item,
+    }
```
leocuisine, line 89: Related to finding #1: the runtime helper builds a fresh `RoutineNS(rtn)` and binds it as `r`, but `OptzBase.e`'s val_map+eval branch never received the same wiring update. If you go with Path A in the top-level review, the simplest fix is `local_vars['rtn'] = self.om.rtn` over in `optzbase.py:326` (and add a fallthrough so `no_parse` rparam-bearing items go through `NumericRoutineNS` instead).

leocuisine (commented):
## Independent review — PR #245 (`eval_e_str` collapse)

I read the new helper, the four refactored opt-layer files, `OptzBase.e`, the symprocessor `val_map` / `sub_map` build, the codegen helpers (`_build_symbol_regex`, `_collect_symbol_names`), `Routine._link_pycode` / `formulation_summary`, and `tests/test_routine_ns.py` + `tests/test_routine.py::test_value_method`. I also reproduced the eval-fallback path locally (DCOPF with `obj.e_str += ' + 0 * cp.sum(pg)'`) to compare behavior against `master`.

The dedup itself is a clear win and the unification of the constraint LHS-vs-Constraint dispatch is good. I found one regression that I think needs a fix before merge, plus a few smaller items.

---

### 1. `correctness-risk` (effectively `bug`) — `OptzBase.e` is broken on the eval-fallback path — `ams/opt/optzbase.py:315`

`self.code` used to be the **post-rewrite** source (`self.om.<name>` references, with `' == 0' / ' <= 0'` appended for Constraints). After this PR `self.code = self.e_str` verbatim. `OptzBase.e` (the post-solve / debug numerical readout, documented in `release-notes.rst:123`) still expects the old shape:

```py
val_map = self.om.rtn.syms.val_map
code = self.code
for pattern, replacement in val_map.items():
    code = re.sub(pattern, replacement, code)
local_vars = {'self': self, 'np': np, 'cp': cp, 'val_map': val_map}
return eval(code, {}, local_vars)
```

`val_map` (built in `ams/core/symprocessor.py:98-194`) substitutes `\b<name>\b → rtn.<name>.v` for parsed symbols, leaves `no_parse` rparams (`c2` etc.) untouched, and strips `(== 0|<= 0)$`. The locals dict has `self`/`np`/`cp` but **no `rtn`** — historically that was fine because `sub_map` had pre-rewritten `pg → self.om.pg`, then `val_map` re-substituted `pg → rtn.pg.v` *inside that string* yielding `self.om.rtn.pg.v` (resolvable through `self`). With the new `self.code = "pg + 1"`, val_map produces `"rtn.pg.v + 1"`, and `eval` raises `NameError: name 'rtn' is not defined`. For `no_parse` rparams the failure is even nastier — `c2` isn't in `val_map` at all, so it stays bare and you get `NameError: name 'c2' is not defined`.

Reproduction (clean checkout of this PR):

```py
ss = ams.load(ams.get_case('5bus/pjm5bus_demo.xlsx'), setup=True, no_output=True, default_config=True)
ss.DCOPF.init()
original = ss.DCOPF.obj.e_str
ss.DCOPF.obj.e_str = original + ' + 0 * cp.sum(pg)'
ss.DCOPF.om.parsed = False; ss.DCOPF.om.evaluated = False
ss.DCOPF.om.parse(); ss.DCOPF.om.evaluate(); ss.DCOPF.om.finalize()
ss.DCOPF.solve(solver='CLARABEL')
print(ss.DCOPF.obj.e)   # → None  (logs "Error in calculating Objective <obj>. name 'c2' is not defined")
```

`obj.v` (the cvxpy-side value) is correct, but `obj.e` silently returns `None` — `OptzBase.e` swallows the exception and logs it. Same regression hits `Constraint.e` and `Expression.e` whenever an item flows through the `sub_map` path (any `addConstrs(...)`, any post-init `e_str = ...` / `e_str +=`, anything `_link_pycode` flagged as `_e_dirty`).

`tests/test_routine.py::test_value_method` doesn't catch this because all of DCOPF's items go through codegen there, so `code is None` and `.e` takes the `e_fn` re-eval branch.

Two ways to restore parity, both fine — pick whichever fits the broader project plan:

- **Path A (smaller diff):** add `'rtn': self.om.rtn` (and probably `'sps': sps`) to the eval namespace in `OptzBase.e`. Then the existing `val_map` substitutions resolve. You'll also want `OptzBase.e` to short-circuit on items where `c2`-style `no_parse` rparams would still be bare — easiest fix is to also fall through to `NumericRoutineNS` whenever `code` is set, mirroring what the e_fn branch does.
- **Path B (cleaner, longer):** change `OptzBase.e` to evaluate via `NumericRoutineNS(self.om.rtn)` for both branches — a "numeric eval_e_str" helper. The PR description says step 4 will introduce `eval_e_str_numeric` and retire `sub_map`/`val_map`; either pull that forward into this PR or land a one-line restore (Path A) now and let step 4 do the real cleanup.

This is the only finding I'd block on; the others are smaller.

---

### 2. `test-gap` — no test exercises `.e` after a customization — `tests/test_routine_ns.py`

`TestRuntimeCustomization.test_e_str_append_then_reinit` is the right place: after `e_str` is mutated and `init()` re-runs, the test checks `exit_code == 0` but never reads `obj.e` or any `constr.e`. Adding one assertion (`assert ss.DCOPF.obj.e is not None and abs(ss.DCOPF.obj.e - ss.DCOPF.obj.v) < 1e-6`) would have caught finding #1 and would catch any future regression in the unified eval path. Same comment applies to the `test_routine.py::test_value_method` loop, which only exercises codegen items.

---

### 3. `nit` — `_collect_symbol_names` re-runs the reserved-name collision check on every `eval_e_str` call — `ams/opt/_runtime_eval.py:57`

`_collect_symbol_names(rtn)` calls `_check_reserved_collisions(...)` on every invocation. The check already runs once in `SymProcessor.generate_symbols` (and once in codegen). For routines with many e_str-fallback items this is repeated sort + set-intersection on every evaluate. Cheap but pure waste. Cache the regex on the routine (e.g. `rtn._symbol_regex`, invalidated when `_syms` flips false), or factor a `_build_symbol_regex_for_routine(rtn)` that memoizes.

---

### 4. `nit` — error-message wrapping uses bare `Exception` — `ams/opt/_runtime_eval.py:69-75`

The old per-call sites also raised bare `Exception`, so this is parity. Worth a follow-up to use a typed `RuntimeError` (or a project-specific `EStrEvaluationError`) so callers can `except` it precisely. Same applies to the bare `Exception` re-raise in `Constraint.evaluate`'s `e_fn` branch.

---

### 5. `nit` — minor behavioral change: e_str containing a relational op is now accepted — `ams/opt/constraint.py:132-136`

Previously `Constraint.parse()` appended `' == 0'` / `' <= 0'` unconditionally to the source string, so an `e_str` like `'pg <= pmax'` was a syntax-level "double op" that would have failed at eval time. With the new `isinstance(result, cp.constraints.Constraint)` check, that same `e_str` now succeeds (the LHS is already a `cp.constraints.Inequality`, so no extra wrap). I think this is a strict improvement, but it is a behavior change worth a one-liner in the release notes.

---

### 6. `style` — `RoutineNS` imported into Constraint/Expression/Objective even though `eval_e_str` already builds one — `ams/opt/constraint.py:15`, `expression.py:11`, `objective.py:13`

Each of these still imports `RoutineNS` at module level for the `e_fn` branch (`self.e_fn(RoutineNS(self.om.rtn))`). Fine, but you now have two construction sites per item per evaluate (one in the e_fn branch in the caller, one inside `eval_e_str` for the e_str branch). Negligible; flagging only because if you eventually fold the e_fn branch through the helper too, the duplicate import goes away.

---

### Things I checked and confirmed are fine

- Symbol-set parity between old `sub_map` keys and `_collect_symbol_names`: vars + rparams + services + exprs + constrs + config keys, with the same `(?<!\.)\b...\b` lookbehind.
- `RoutineNS` resolution table matches what the old `sub_map` rewrote to (vars→`optz`, rparams→matprocessor / `om` / `.v`, services→`om` / `.v`, exprs→`om`, constrs→`optz`, config→`config.<key>`).
- Removal of `import numpy as np` from `objective.py` is safe — there were no `np.` references in that file's e_str-eval block (the old block only had `'cp': cp` in locals); other files retained `numpy` because they reference `np.ndarray` in setters.
- Constraint LHS-vs-Constraint unification: when the old e_fn branch returned a bare LHS, it was wrapped via `is_eq`; the old e_str branch returned a Constraint (because `' == 0' / ' <= 0'` was appended). New code dispatches on `isinstance(result, cp.constraints.Constraint)` for both — semantically equivalent for the common case and strictly more permissive for the corner case in finding #5.
- `formulation_summary` reads `_e_str` directly, not `code` — unaffected by the new `self.code` semantics.
- `# nosec B307` rationale is clear and matches the existing trust boundary in codegen.
- `_e_dirty` propagation through the descriptor mutex still works for the `e_str += ...` pattern.

ams/opt/constraint.py
leocuisine: Fixed in cad59ff8e6e233c95eca8b903e31305a4f870d03 via Path B from your suggestion — added `eval_e_str_numeric` in `_runtime_eval.py` and rewired `OptzBase.e`'s e_str branch through it. Reproduced your original failure (`name 'c2' is not defined`) on the prior commit; `obj.e == obj.v` now after the fix.

ams/opt/_runtime_eval.py
leocuisine: Done in cad59ff8e6e233c95eca8b903e31305a4f870d03. `OptzBase.e` now delegates to `eval_e_str_numeric` (NumericRoutineNS-backed). The CVXPY-side helper still binds a fresh `RoutineNS`; refactoring the e_fn branches through the helper is left as a follow-up (see Steps 3–5 of the project plan).

leocuisine:
Thanks for the careful review. All six findings addressed in cad59ff8:

1. **`correctness-risk` (`OptzBase.e` broken on eval-fallback)** — fixed via your suggested Path B. Added `eval_e_str_numeric(item, e_str)` in `ams/opt/_runtime_eval.py` (NumericRoutineNS-backed, mirrors the CVXPY-side helper's symbol-resolution surface), rewired `OptzBase.e`'s e_str branch to it, dropped the val_map+eval block. Reproduced your `name 'c2' is not defined` failure on the prior commit; verified the fix.

2. **`test-gap` (no `.e` assertion after customization)** — added in `tests/test_routine_ns.py::TestRuntimeCustomization.test_e_str_append_then_reinit`. The new asserts (`obj.e is not None` + `obj.e ≈ obj.v`) reproduce finding #1 on the broken commit and pass after the fix.

3. **`nit` (regex re-built per call)** — cached on `rtn.syms._eval_symbol_regex` via `_get_symbol_regex(rtn)`. Invalidated in `SymProcessor.generate_symbols` when symbols regenerate.

4. **`nit` (bare `Exception` re-raise)** — acknowledged but deferring; matches the legacy parity per your own note. Tracked for the typed-exception sweep that's already in the architecture review's TODO inventory.

5. **`nit` (Constraint behavioral change re: embedded relational ops)** — added a paragraph to the v1.2.2 release-notes "Migration" section explaining the new dispatch-on-isinstance behavior.

6. **`style` (RoutineNS imported in caller files)** — acknowledged; deferring to the future PR that folds the `e_fn` branch through the helper too (will eliminate the duplicate construction).

Also dropped now-unused imports from `optzbase.py` (`re`, `np`, `pretty_long_message`, `_prefix`, `_max_length`).

Full pytest tests/ — 348 passed; flake8 clean.

PR #246, opened 2026-05-02
refactor(opt): retire SymProcessor.sub_map and val_map (cvxpy-ns Step 4)

leocuisine:
## Summary

Step 4 of the cvxpy-namespace passthrough project. Both `SymProcessor.sub_map` and `SymProcessor.val_map` are populated-but-dead after PR #245 — Steps 2/3 rerouted every reader through `ams.opt._runtime_eval` + `RoutineNS` / `NumericRoutineNS`. This PR deletes the structures plus the per-symbol writes inside `generate_symbols()`.

- `sub_map`'s symbol substitutions are now produced on demand by `_runtime_eval._resolve` via `_build_symbol_regex` against the routine's symbol registries.
- `val_map`'s numpy rewrites are now produced by `NumericRoutineNS` attribute access through `eval_e_str_numeric`.

The unused `MatProcessor` import (only the rparam dispatch needed it) is dropped. The `'sub_map'` *string label* in `formulation_source` and the corresponding tally keys in `routine.py` / `optzbase.py` are deliberately untouched — those are Step 5's rename to `'eval'`.

## Diff stats

```
 ams/core/routine_ns.py        |  5 +--
 ams/core/symprocessor.py      | 72 +++----------------------------------------
 docs/source/release-notes.rst | 10 ++++++
 tests/test_routine_ns.py      |  8 ++---
 4 files changed, 21 insertions(+), 74 deletions(-)
```

## Test plan

- [x] `pytest tests/` — 348 passed locally on macOS / Python 3.12 / CLARABEL.
- [x] `flake8 ams/ --max-line-length=120 --exclude=ams/thirdparty` — clean.
- [x] DCOPF: `obj.e == obj.v` on both the codegen path and the eval-fallback path after `obj.e_str += ' + 0'` + reparse.
- [x] `hasattr(ss.DCOPF.syms, 'sub_map')` and `hasattr(ss.DCOPF.syms, 'val_map')` both `False` after `init()`.
- [ ] CI green on this branch.

### DIFF
--- a/ams/core/routine_ns.py
+++ b/ams/core/routine_ns.py
@@ -46,8 +46,9 @@ class RoutineNS:
     Attribute proxy passed as ``r`` to ``e_fn(r)`` callables.
 
     Holds a reference to the owning routine and resolves attribute
-    access by walking the same five symbol buckets the legacy
-    ``sub_map`` regex used. Read-only — there is no setter.
+    access by walking the five routine symbol buckets (vars, rparams,
+    services, exprs, constrs) plus config and a couple of
+    BaseRoutine conventions. Read-only — there is no setter.
     """
 
     __slots__ = ('_rtn',)
--- a/ams/core/symprocessor.py
+++ b/ams/core/symprocessor.py
@@ -16,7 +16,6 @@ from collections import OrderedDict
 import sympy as sp
 
 from ams.utils.misc import elapsed
-from ams.core.matprocessor import MatProcessor
 
 logger = logging.getLogger(__name__)
 
@@ -32,12 +31,8 @@ class SymProcessor:
 
     Attributes
     ----------
-    sub_map : dict
-        Substitution map for symbolic processing.
     tex_map : dict
         Tex substitution map for documentation.
-    val_map : dict
-        Value substitution map for post-solving value evaluation.
     """
 
     def __init__(self, parent):
@@ -47,19 +42,6 @@ class SymProcessor:
         self.config = parent.config
         self.class_name = parent.class_name
         self.tex_names = OrderedDict()
-        self.tex_map = OrderedDict()
-
-        # CVXPY is the sole modeling language. As of the
-        # cvxpy-namespace passthrough (Step 2), routines author
-        # canonical CVXPY directly (``cp.sum(...)``, ``cp.multiply(...)``,
-        # ``*`` for element-wise). The function-name rewrite block
-        # that previously lived here (``mul → cp.multiply``,
-        # ``dot → *`` etc.) is gone; ``sub_map`` now carries only the
-        # per-symbol substitutions populated by ``generate_symbols``.
-        # The legacy eval-fallback path in
-        # ``opt/{constraint,objective,expression,exprcalc}.py`` therefore
-        # requires its inputs to already be in canonical CVXPY form.
-        self.sub_map = OrderedDict()
 
         # First rule strips the ``cp.`` Python-module prefix from
         # canonical-CVXPY e_str (added in the namespace-passthrough
@@ -92,16 +74,6 @@ class SymProcessor:
             (r'\bsum\b', 'SUM'),
             (r'power\((.*?),\s*(\d+)\)', r'\1^\2'),
             (r'(\w+).dual_variables\[0\]', r'\phi[\1]'),
-        ])
-
-        # mapping dict for evaluating expressions
-        self.val_map = OrderedDict([
-            (r'(== 0|<= 0)$', ''),  # remove the comparison operator
-            (r'cp\.(Minimize|Maximize)', r'float'),  # remove cp.Minimize/Maximize
-            (r'\bcp.\b', 'np.'),
-            (r'\bexp\b', 'np.exp'),
-            (r'\blog\b', 'np.log'),
-            (r'\bconj\b', 'np.conj'),
         ])
 
         self.status = {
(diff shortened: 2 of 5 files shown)


leocuisine (commented):
**Verdict:** LGTM — approve with two minor doc/comment nits.

I verified the deletion is complete and safe end-to-end. Step 4's contract — kill `sub_map` / `val_map` as attributes while preserving everything `generate_symbols()` needs to keep producing — is met.

### What I checked

1. **No remaining attribute readers.** Repo-wide sweep for `syms.sub_map`, `syms.val_map`, `.sub_map[...]`, `.val_map[...]`, `.sub_map.get(...)`, `.val_map.get(...)` returns nothing. The remaining `'sub_map'` references are exclusively the `formulation_source` string label in `ams/opt/optzbase.py:159,199-215` and the tally keys in `ams/routines/routine.py:289-356` — both explicitly Step 5 territory, plus historical-context comments and example notebook output.
2. **`generate_symbols()` post-edit correctness.** All four downstream-consumed structures still get populated: `inputs_dict`, `services_dict`, `tex_names`, `tex_map`. Verified `tex_map` consumer at `ams/core/documenter.py:359,414,467,507` (LaTeX rendering); `inputs_dict` consumer at `_check_expr_symbols` (line 184). The five-bucket walk pattern for the reserved-name set matches the categories now sourced for `inputs_dict` / `tex_map`.
3. **Eval-fallback intact.** `eval_e_str` and `eval_e_str_numeric` (`ams/opt/_runtime_eval.py`) reach for `_collect_symbol_names` + `_build_symbol_regex` from `ams.prep.generator` — neither touches the deleted dicts. `RoutineNS` / `NumericRoutineNS` cover all six resolution categories (vars / rparams / services / exprs / constrs / config) plus `sys_f` / `sys_mva`, matching exactly what `sub_map` and `val_map` used to encode.
4. **Reserved-name check still runs at the right spot.** `_check_reserved_collisions` now executes at `ams/core/symprocessor.py:116-122`, *before* any state mutation, against the union of `vars`/`rparams`/`services`/`exprs`/`constrs`. Same five categories the old per-bucket loops walked. Failure raises before any partial population — clean.
5. **`MatProcessor` import removal safe.** Grep confirms no other reach in `ams/core/symprocessor.py`. The MatProcessor `isinstance` dispatch moved to `RoutineNS._is_matprocessor` (lazy import, already in place from Step 3).
6. **`_eval_symbol_regex` cache invalidation preserved** at line 108, so re-`generate_symbols()` after `addRParam` / `addConstrs` still reseeds the cached regex on next eval.
7. **Test coverage is adequate.** `tests/test_routine_ns.py::TestRoutineNSResolution` (6 tests) exercises the resolution surface end-to-end on DCOPF, and `TestRuntimeCustomization::test_e_str_append_then_reinit` (lines 164-198) hits the exact eval-fallback path — including `obj.e ≈ obj.v`, which is the regression net for `eval_e_str_numeric`. Anything calling `syms.sub_map['...']` would NameError immediately on import / init, so a bad regression here would be loud, not silent.
8. **Release notes** correctly call out the attribute removal under v1.2.2 Migration with a redirect to `formulation_source` + the routine registries.

### Findings

1. **`ams/prep/generator.py:41-49,64` — stale `sub_map` reference in `RESERVED_CVXPY_ATOM_NAMES` rationale** — *nit (docs)*. The block comment ("``SymProcessor`` writes a ``sub_map`` rule like…") and the `ValueError` message inside `_check_reserved_collisions` ("eval-fallback ``sub_map`` would silently rewrite…") both still describe the deleted mechanism. The reserved-name guard is still load-bearing — it now protects the `_runtime_eval` path's `_resolve()` regex rewrite to `r.<name>` — but the rationale text is one rename behind. Suggest swapping `sub_map` → `_runtime_eval` (or "the eval-fallback rewrite") in both spots so an end user hitting the error sees an accurate explanation. Worth folding into Step 5's rename PR if not before.

2. **`ams/opt/optzbase.py:159` — comment references `sub_map+eval path`** — *nit (docs)*. Also Step 5 territory, but flagging here in case Step 5's scope is read narrowly as "rename the string label only" rather than "rename the concept." This single-line comment about provenance values would otherwise outlive the rename.

### Suggested pre-merge sanity check

Already covered locally per your test plan, but worth re-running once on CI: `pytest tests/test_routine_ns.py -k "test_e_str_append_then_reinit"` is the single test that, if it goes red, indicates a real eval-fallback regression introduced by Step 4. Everything else is a tighter loop on the same surface.

leocuisine:
Thanks for the review. Disposition on both nits:

**Nit #1 (`generator.py` reserved-name comment + ValueError) — fixed in `440565bd`.** Both the `RESERVED_CVXPY_ATOM_NAMES` rationale block and the `_check_reserved_collisions` ValueError now point at `ams.opt._runtime_eval` instead of the deleted `sub_map`. The error message a tripping user sees is the user-facing surface, so worth getting right in this PR.

**Nit #2 (`optzbase.py:159` comment about `sub_map+eval path`) — deferring to Step 5.** That comment lives on the `_e_fn_source` declaration and lines up with `formulation_source` literally returning the string `'sub_map'` until Step 5's bucket rename. Touching the comment now without renaming the bucket would make the comment talk about a value that doesn't appear in the same file. Step 5 will rewrite both at once.

Local checks after `440565bd`: 348 tests pass, flake8 clean.

PR #249, opened 2026-05-02
refactor(opt): retire Constraint.is_eq — embedded-operator LHS-zero e_str

leocuisine:
## Summary

Closes the cvxpy-namespace passthrough follow-on (`projects/is_eq_retirement/`).

Drops the `is_eq` parameter from `Constraint.__init__` and `RoutineBase.addConstrs`. Every Constraint now embeds the relational operator directly in `e_str` — `'<LHS> <= 0'`, `'<LHS> == 0'`, or `'<LHS> >= 0'` — preserving the LHS-zero authoring discipline so `Constraint.v` keeps reporting **slack-from-zero** uniformly.

**Why this works:** CVXPY canonicalizes every inequality to `_expr = lhs - rhs` regardless of `<=` / `>=` direction (verified — `(b >= a)._expr` is `a - b`, not `b - a`). The diagnostic invariant rides on `_expr`; both authoring forms preserve it. With `is_eq` gone, the operator lives in exactly one place (the `e_str` text), eliminating the silent-override footgun (`e_str='pg <= pmax', is_eq=True` previously ignored `is_eq`).

## Architectural decision: single source of truth in `e_str`

`documenter.py:360` previously branched on `p.is_eq` to render `= 0` vs `\leq 0`. The instinct was to stash a derived `_is_eq` from parsing `e_str` (option b in research) — but that re-creates the parallel-truth channel as a private attribute. Chose option (c): three new `tex_map` / `_TEX_TEMPLATES` rules (`<= -> \leq`, `>= -> \geq`, `== -> =`) translate the embedded operator into LaTeX. Documenter loop collapses to one `_tex_pre()` call. Single source of truth.

## What changed

**Core:**
- `Constraint.__init__` drops `is_eq` parameter and `self.is_eq` attribute.
- `Constraint.evaluate()` requires `cp.constraints.Constraint` result; raises `TypeError` with a clear "embed `<= 0` / `== 0` / `>= 0`" message otherwise.
- `RoutineBase.addConstrs` drops `is_eq` parameter; refreshed docstring + example.
- `eval_e_str_numeric` strips the trailing operator before numpy eval (else short-circuits to a bool array).
- `OptzBase.e` reorders to prefer the e_str-driven numeric path whenever `e_str` is available (preserved post-codegen wiring) — uniform for codegen and eval-fallback paths.
- `documenter.py:360-364` simplified — branching deleted.
- Three relational tex rules added to both `SymProcessor.tex_map` (`ams/core/symprocessor.py`) and `_TEX_TEMPLATES` (`ams/prep/generator.py`) — they're documented to "both must agree".

**Routine sites:** ~73 Constraint constructors across 12 files (`dcopf.py`, `dcopf2.py`, `dcpf.py`, `dopf.py`, `ed.py`, `ed2.py`, `pflow.py`, `pypower.py`, `routine.py`, `rted.py`, `uc.py`, `uc2.py`). Mechanical migration via `icebar/migrate_is_eq.py` (gitignored), plus hand-pass for ~16 deferred-assignment sites where `is_eq` lived in a sibling declaration.

**Docs:**
- `docs/source/release-notes.rst` — new v1.2.2 migration section.
- `docs/source/modeling/migration_cvxpy_namespace.rst` — new "is_eq retirement" section with rationale, before/after examples, custom-callable note.
- `docs/source/modeling/routine.rst` — refresh `is_eq` mention, sample Constraint snippet.
- `docs/source/modeling/example.rst` — DCOPF example updated.

**Notebooks:**
- `examples/ex8.ipynb` — `addConstrs` calls migrated, comment explaining the move.
- `examples/demonstration/demo_debug.ipynb` — new explainer cell on CVXPY's canonicalization (per maintainer's 2026-05-02 question that motivated this PR), making the `.e` slack diagnostic feel less mysterious.

## Test plan

- [x] `pytest tests/`: 348 passed.
- [x] `flake8 ams/ --max-line-length=120 --exclude=ams/thirdparty`: clean.
- [x] `make distclean html SPHINXOPTS='-W --keep-going -j auto'`: clean.
- [x] `jupyter nbconvert --execute --inplace ex8.ipynb demo_debug.ipynb`: re-executed cleanly with refreshed outputs.
- [x] Generated pycode tex strings render correctly: `pglb -> '... \\leq 0'`, `pb -> '... = 0'`.
- [ ] Reviewer agent — independent reviewer for accuracy of routine-site migrations + correctness of `.e` codegen-path reorder.

### DIFF
--- a/ams/core/documenter.py
+++ b/ams/core/documenter.py
@@ -352,15 +352,15 @@ class RDocumenter:
             class_names.append(p.class_name)
             info.append(p.info if p.info else '')
 
-        # expressions based on output format
+        # expressions based on output format. The relational operator
+        # is now part of ``e_str`` itself (post `is_eq` retirement);
+        # ``_tex_pre`` translates ``<= 0`` / ``== 0`` / ``>= 0`` to
+        # the LaTeX form via the trio of rules at the tail of
+        # ``_TEX_TEMPLATES`` (and the parallel ``SymProcessor.tex_map``).
         expressions = []
         if export == 'rest':
             for p in self.parent.constrs.values():
                 expr = _tex_pre(self, p, self.parent.syms.tex_map)
-                if p.is_eq:
-                    expr = f'{expr} = 0'
-                else:
-                    expr = f'{expr} \\leq 0'
                 logger.debug(f'{p.name} math: {expr}')
                 expressions.append(expr)
 
--- a/ams/core/symprocessor.py
+++ b/ams/core/symprocessor.py
@@ -74,6 +74,15 @@ class SymProcessor:
             (r'\bsum\b', 'SUM'),
             (r'power\((.*?),\s*(\d+)\)', r'\1^\2'),
             (r'(\w+).dual_variables\[0\]', r'\phi[\1]'),
+            # Relational operators in embedded-operator e_str (post
+            # `is_eq` retirement). Ordering within this trio doesn't
+            # matter (no overlap among `<=`, `>=`, `==`). Double
+            # backslash in the replacement so `re.sub` emits a
+            # literal `\leq` / `\geq` (Python 3.12+ raises
+            # re.PatternError on the single-backslash form).
+            (r'<=', r'\\leq'),
+            (r'>=', r'\\geq'),
+            (r'==', '='),
         ])
 
         self.status = {
--- a/ams/opt/_runtime_eval.py
+++ b/ams/opt/_runtime_eval.py
@@ -20,6 +20,7 @@ function-rewrite layer was deleted in PR #244):
 """
 
 import logging
+import re
 
 import cvxpy as cp
 import numpy as np
@@ -102,6 +103,9 @@ def eval_e_str(item, e_str):
         ) from exc
 
 
+_TRAILING_OP_RE = re.compile(r'\s*(==|<=|>=)\s*0\s*$')
+
+
 def eval_e_str_numeric(item, e_str):
     """Evaluate ``e_str`` numerically for :pyattr:`OptzBase.e`.
 
@@ -114,12 +118,16 @@ def eval_e_str_numeric(item, e_str):
     ``val_map`` + ``np.<atom>`` rewrite produced, but driven by the
     same name-resolution surface as the CVXPY-side helper.
 
-    For a ``cp.constraints.Constraint`` result (e.g. an ``e_str``
-    that already contains ``<= 0``), returns the constraint's LHS
-    expression so callers can ``.value`` it uniformly.
+    For an ``e_str`` carrying an embedded relational operator
+    (``... <= 0`` / ``... == 0`` / ``... >= 0``), strip the trailing
+    operator before evaluating so the returned object is the LHS
+    slack — what ``Constraint.v`` reports via CVXPY canonicalization
+    on the symbolic side. Without the strip, a numpy-only eval
+    short-circuits to a bool array.
     """
     rtn = item.om.rtn
-    code = _resolve(rtn, e_str)
+    body = _TRAILING_OP_RE.sub('', e_str)
+    code = _resolve(rtn, body)
     local_vars = {
         'cp': cp, 'np': np, 'sps': sps,
         'r': NumericRoutineNS(rtn), 'self': item,
--- a/ams/opt/constraint.py
+++ b/ams/opt/constraint.py
@@ -134,11 +134,16 @@ class Constraint(OptzBase):
             logger.debug(pretty_long_message(msg, _prefix, max_length=_max_length))
             result = eval_e_str(self, self.e_str)
         if not isinstance(result, cp.constraints.Constraint):
+            source = f"e_str={self.e_str!r}" if self.e_str is not None else "e_fn"
             raise TypeError(
-                f"Constraint <{self.name}> e_str/e_fn must produce a "
+                f"Constraint <{self.name}>: {source} must produce a "
                 f"cp.constraints.Constraint (embed the relational "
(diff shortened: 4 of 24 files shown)


leocuisine (commented):
## Independent reviewer agent — PR #249

Sample-checked all routine-site migrations against `git show develop:<file>`, exercised the codegen pycode for DCOPF (tex strings render `\leq 0` / `= 0` correctly), and ran the full test suite (348 tests). The two failures observed (`tests/test_group.py::TestGroup::test_group_access`, `tests/test_interface.py::TestDataExchange::test_data_exchange`) are **pre-existing on master** — neither test file nor the modules they touch (`ams/models/group.py`, `ams/interface.py`) appears in this PR's diff, and the failures are signature/method-name issues unrelated to constraint refactoring.

The architectural decision (single source of truth in `e_str`, three relational rules in `tex_map` / `_TEX_TEMPLATES`) is clean. `is_eq` polarities all preserved correctly under the `True → == 0` / `False → <= 0` mapping. Only minor findings below.

---

### Findings

**1. `style` — `ams/opt/_runtime_eval.py:106` — `_TRAILING_OP_RE` could be docstring-anchored to its constraint-only contract.**

The regex `r'\s*(==|<=|>=)\s*0\s*$'` is correct for the single intended caller (`eval_e_str_numeric` on a `Constraint.e_str`), but `eval_e_str_numeric` is also reachable from `OptzBase.e` for **Objective**, **Expression**, **ExpressionCalc** items via the reordered property. For those item types the e_str doesn't end in a relational op, so the strip is a no-op — but a one-liner near the regex noting "constraint-only intent; no-op for objectives/expressions because their `e_str` never carries a trailing relational op" would document the cross-class safety. Not a behavior issue; the no-op outcome is exactly what's needed.

**2. `nit` — `ams/opt/optzbase.py:285-313` — duplicate cp.Constraint LHS-extraction logic with `eval_e_str_numeric`.**

The e_fn-only fallback in `OptzBase.e` (lines 305-310) extracts the LHS via `_expr` / `args[0]` exactly the way `eval_e_str_numeric` (lines 145-149 of `_runtime_eval.py`) does. Pulling the two-line extraction into a small helper (`_extract_constr_lhs(result)`) would dedupe and make the parallel intent explicit. Not blocking — both copies are correct.

**3. `nit` — `ams/opt/constraint.py:136-142` — TypeError message could name the actual class.**

The new error message is clear and actionable. The wording `"Got {type(result).__name__}"` is helpful; one extra signal that would help a debugger is to also surface `self.e_str` (or `self.e_fn.__name__` for the e_fn path) so the user sees the exact authoring text without having to introspect — useful when many Constraints fan out from a loop. Optional polish.

**4. `nit` — `ams/opt/_runtime_eval.py:121-126` — docstring "what ``Constraint.v`` reports" is slightly inaccurate for non-Constraint callers.**

The docstring frames the trailing-op strip as "what ``Constraint.v`` reports via CVXPY canonicalization on the symbolic side." Since this helper now serves Objective and Expression as well (per the `OptzBase.e` reorder), the framing is constraint-specific. Consider a parenthetical: "(no-op for Objective/Expression e_strs, which don't carry a trailing relational op)."

**5. `nit` — `docs/source/release-notes.rst:667` — historical "is_eq" reference in old changelog.**

Existing entry: `Rename ``type='eq'`` to ``is_eq=False`` in ``Constraint`` to avoid overriding built-in attribute`. This is a historical changelog entry, so leaving it is correct (changelog == time-locked record). Calling out only because the search for stale references caught it; **no action needed**.

**6. `nit` — pycode regeneration in CI / installed envs.**

After this merge, any user with stale pycode in `~/.ams/pycode/` (containing the pre-`is_eq`-retirement `_constr_*` callables that returned LHS-only) will see `TypeError` from the new `Constraint.evaluate()` because the cached callables don't return a `cp.constraints.Constraint`. This is the correct behavior (the new contract requires it), but a one-line note in the migration doc — "delete `~/.ams/pycode/` after upgrade if you have any cached pycode older than v1.2.2" — would save users a confusing error. The `pycode` regeneration is automatic on next `init()` once the cache is cleared, but the cached files don't auto-invalidate on package upgrade.

**7. `nit` — `ams/core/symprocessor.py:83-85` and `ams/prep/generator.py:115-117` — duplicated rule trio.**

The "both must agree" comment is accurate and the duplication is intentional (per the file-level docstring at `_TEX_TEMPLATES`). Not actionable in this PR; just noting that any future rule additions need to update both places.

---

LGTM otherwise. The migration is mechanical, the operator embedding makes the source-of-truth story crisper, and the `OptzBase.e` reorder correctly preserves diagnostic semantics across all four item types.

leocuisine:
Thanks — addressed in `3294326d`:

- **#6 (cached pycode)**: bumped `PYCODE_FORMAT_VERSION` 2→3. The existing `_link_pycode` mismatch path already auto-invalidates caches whose version doesn't match, so users hit the new TypeError only if they freshly re-author an `e_str` without the trailing operator — not from stale upgrade state. No manual `ams prep --force` needed.
- **#3 (TypeError context)**: `Constraint.evaluate`'s `TypeError` now surfaces `self.e_str` (or "e_fn" when none) and points at the auto-invalidation behavior so anyone seeing the error from a freshly-regenerated cache knows the underlying `e_str` is missing its trailing operator.
- Migration guide gains a one-paragraph "Cached pycode is auto-invalidated" note.

Skipping the cosmetic refinements (#1/#2/#4 dedup + docstring framing, #5 historical changelog, #7 documented-mirror trio) — happy to revisit any of them in a follow-up if you'd prefer.

`pytest tests/`: 348 passed. `make html` (-W --keep-going): clean.

PR #250, opened 2026-05-02
test: add tests/conftest.py with shared case fixtures (Phase 1 of tests refactor)

leocuisine:
## Summary

Phase 1 of the **tests/ refactorization** project ([plan](../blob/refactor/tests-conftest-phase1/.claude/projects/tests_refactorization/plan.md), local). Lays the fixture plumbing so later phases (scenario consolidation, known-good parametrization, layer rename) have a single place to plug in. **No structural change yet** — only `conftest.py` plus opportunistic adoption in two files that were trivially compatible.

### What changes
- **New `tests/conftest.py`** — seven session-scoped raw-load fixtures (`pjm5bus_json_loaded`, `pjm5bus_xlsx_loaded`, `case5_loaded`, `case14_loaded`, `ieee14_raw_loaded`, `ieee39_uced_loaded`, `ieee39_uced_esd1_loaded`) and function-scoped deepcopy wrappers without the `_loaded` suffix. Each top case is loaded **once per session, lazily**, and tests get a fresh deepcopy so mutations stay isolated.
- **`tests/test_omodel.py`** — 3 classes migrated to `@pytest.fixture(autouse=True)` injecting `self.ss` from `case5` / `pjm5bus_json` fixtures.
- **`tests/test_routine.py`** — 3 classes migrated similarly (`ieee39_uced` and `pjm5bus_json`).

### Why deepcopy and not `ss.reset()`
Verified empirically (research notes in plan.md):
- `ss.reset()` does **not** undo `PQ.p0.v`, `StaticGen.u`, or `routine.initialized` mutations.
- `ams.load(5bus)` ≈ 80 ms; `copy.deepcopy(ss)` ≈ 60 ms; `ss.reset()` ≈ 1.2 ms but unsafe for shared fixtures.

Per-test deepcopy is the only safe per-test isolation. The session-scoped raw-load fixtures (`*_loaded`) are exposed for read-only consumers that want to skip the deepcopy cost.

### What's intentionally left alone
- All 11 `test_rtn_*.py` files — these get rewritten wholesale in **Phase 2** (parametrized scenario consolidation). Touching them now would create churn.
- `test_routine_ns.py` — already uses `setUpClass` with a read-only shared `cls.ss`, which is optimal. Converting to a class-scope fixture would be scope creep with no measurable win.
- Layer rename / subfolder reorganization — that's Phase 4.

### Wall-time

| Metric | Baseline (develop) | Phase 1 | Delta |
|---|---|---|---|
| Tests collected | 348 | 348 | — |
| Wall time (`pytest tests/ -q`) | 61.80 s | **52.27 s** | **−9.5 s (−15 %)** |
| Slowest UC/ED `set_load` tests | 1.0–1.3 s | 0.92–1.07 s | small |

The win is larger than the conservative ~2 s estimate because the migrated `TestOModelrepr` / `TestRoutineMethods` classes no longer pay the `setUp()` re-load cost on every method (the bigger `ieee39_uced` case is loaded once instead of 7 times).

### Test plan
- [x] `pytest tests/ -q --durations=20` — 348 passed, 0 failed, 9.5 s faster than baseline.
- [x] `flake8 ams/ tests/` — clean.
- [x] `pytest tests/test_omodel.py tests/test_routine.py -v` — all 19 migrated tests green.
- [ ] Independent reviewer agent (to be spawned post-PR-open).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,67 @@
+"""
+Shared pytest fixtures for the AMS test suite.
+
+Each fixture loads its case once per pytest session (lazy, on first
+request) and yields a fresh deepcopy on every test invocation, so
+tests that mutate ``ss`` (parameters, routine state, ``StaticGen.u``)
+stay isolated. ``ss.reset()`` is *not* sufficient — it does not undo
+parameter mutations or ``routine.initialized``.
+
+Usage from an existing ``unittest.TestCase``::
+
+    class TestFoo(unittest.TestCase):
+        @pytest.fixture(autouse=True)
+        def _attach_ss(self, pjm5bus_json):
+            self.ss = pjm5bus_json
+
+Or directly from a pytest-style function::
+
+    def test_bar(pjm5bus_json):
+        assert pjm5bus_json.DCOPF is not None
+
+The session-scoped raw-load fixtures (``*_loaded``) are exposed for
+read-only consumers that want to skip the deepcopy cost.
+"""
+
+import copy
+
+import pytest
+
+import ams
+
+
+def _make_loaded_fixture(case_path):
+    @pytest.fixture(scope="session")
+    def _loaded():
+        return ams.load(
+            ams.get_case(case_path),
+            setup=True,
+            default_config=True,
+            no_output=True,
+        )
+    return _loaded
+
+
+def _make_fresh_fixture(loaded_name):
+    @pytest.fixture
+    def _fresh(request):
+        return copy.deepcopy(request.getfixturevalue(loaded_name))
+    return _fresh
+
+
+pjm5bus_json_loaded = _make_loaded_fixture("5bus/pjm5bus_demo.json")
+pjm5bus_xlsx_loaded = _make_loaded_fixture("5bus/pjm5bus_demo.xlsx")
+case5_loaded = _make_loaded_fixture("matpower/case5.m")
+case14_loaded = _make_loaded_fixture("matpower/case14.m")
+ieee14_raw_loaded = _make_loaded_fixture("ieee14/ieee14.raw")
+ieee39_uced_loaded = _make_loaded_fixture("ieee39/ieee39_uced.xlsx")
+ieee39_uced_esd1_loaded = _make_loaded_fixture("ieee39/ieee39_uced_esd1.xlsx")
+
+
+pjm5bus_json = _make_fresh_fixture("pjm5bus_json_loaded")
+pjm5bus_xlsx = _make_fresh_fixture("pjm5bus_xlsx_loaded")
+case5 = _make_fresh_fixture("case5_loaded")
+case14 = _make_fresh_fixture("case14_loaded")
+ieee14_raw = _make_fresh_fixture("ieee14_raw_loaded")
+ieee39_uced = _make_fresh_fixture("ieee39_uced_loaded")
+ieee39_uced_esd1 = _make_fresh_fixture("ieee39_uced_esd1_loaded")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,26 +1,31 @@
 """
 Shared pytest fixtures for the AMS test suite.
 
-Each fixture loads its case once per pytest session (lazy, on first
-request) and yields a fresh deepcopy on every test invocation, so
-tests that mutate ``ss`` (parameters, routine state, ``StaticGen.u``)
-stay isolated. ``ss.reset()`` is *not* sufficient — it does not undo
-parameter mutations or ``routine.initialized``.
+Each fixture loads its case **once per worker process** (lazy, on
+first request, cached in module-level state) and yields a fresh
+``copy.deepcopy`` on every test invocation, so tests that mutate
+``ss`` (parameters, routine state, ``StaticGen.u``) stay isolated.
+``System.reset()`` is *not* sufficient — it does not undo parameter
+mutations or ``routine.initialized``.
 
 Usage from an existing ``unittest.TestCase``::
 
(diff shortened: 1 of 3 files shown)


leocuisine (commented):
Independent reviewer pass on PR #250 (fresh eyes, no prior context). Verified locally: full collection (`pytest tests/ --co`) yields 348 tests; the 19 migrated tests pass; `--fixtures` listing shows all 14 fixtures register correctly under their module-binding names.

Net assessment: the design is sound. The correctness arguments for deepcopy-per-test (over `ss.reset()`) hold up — `System.reset()` at `ams/system.py:382` only re-runs `setup()`, which won't restore mutated `PQ.p0.v` / `StaticGen.u` or clear `routine.initialized`. Session-scope + per-test deepcopy is the right shape.

A few findings, none blocking.

---

### 1. `[correctness-risk]` `tests/conftest.py:33-42` — session-scoped `*_loaded` fixtures are shared mutable state with only doc-level guardrails

`_make_loaded_fixture` returns a session-scoped `System` with no copy-on-read protection. Any future test that requests `pjm5bus_json_loaded` directly (the docstring at lines 22-23 advertises this for "read-only consumers") and mutates it — even unintentionally, e.g. by calling `.run()`, which sets `routine.initialized` and writes `pg.v` — will silently corrupt every subsequent test in the session that depends on the same fixture. Failures will manifest as order-dependent flakiness, the worst kind to debug.

The docstring warning is the only safeguard. Two cheap hardenings to consider (either, not both):

- Don't expose `*_loaded` publicly at all in Phase 1. Keep them as the private upstream of `_fresh` and add them later when an actual read-only consumer materializes. YAGNI.
- Or, when a `_loaded` fixture is requested, attach a tripwire: e.g. snapshot a hash of a few well-known fields in a session-finalizer and assert unchanged. Cheap, catches the regression class.

Not blocking — Phase 1 doesn't yet have any direct `_loaded` consumers — but worth resolving before Phase 2 starts wiring them in.

### 2. `[style]` `tests/conftest.py:34-49` — factory-produced fixtures all share `__name__` (`_loaded`, `_fresh`), hurting introspection

```
$ pytest tests/test_omodel.py --fixtures -q
case14_loaded [session scope] -- tests/conftest.py:35
case5_loaded  [session scope] -- tests/conftest.py:35
ieee14_raw_loaded [session scope] -- tests/conftest.py:35
...
```

All seven `_loaded` fixtures point at line 35 and all seven `_fresh` at line 47, because the closure carries `__name__ = '_loaded'` / `'_fresh'`. Pytest registers under the module-binding name so functionality is fine, but tracebacks, `--fixtures` listing, and any third-party pytest plugin that introspects `fixture.__name__` will see the generic name. Trivial fix inside each factory:

```python
def _make_loaded_fixture(case_path):
    @pytest.fixture(scope="session")
    def _loaded():
        return ams.load(ams.get_case(case_path), setup=True,
                        default_config=True, no_output=True)
    _loaded.__doc__ = f"Session-scoped raw load of {case_path}."
    return _loaded
```

…and then assign `__name__` at module-binding time:

```python
pjm5bus_json_loaded = _make_loaded_fixture("5bus/pjm5bus_demo.json")
pjm5bus_json_loaded.__name__ = "pjm5bus_json_loaded"
```

Or skip the factory and write seven explicit fixtures — at seven cases the DRY win is marginal vs. the diagnostics loss.

### 3. `[nit]` `tests/test_routine.py:1-6` — import grouping is mildly non-PEP8

```
import unittest
import numpy as np

import pytest

from ams.shared import pd
```

`numpy` and `pytest` are both third-party; PEP 8 wants them in one block, then a blank line, then first-party (`ams.shared`). flake8 doesn't enforce this so it's clean per CI, but `isort` would reorder. Low priority; mention only because conftest.py introduction is a natural moment to settle the convention for the suite.

### 4. `[nit]` `tests/conftest.py:36-41` and migrated `setUp` removals — load kwargs no longer locally documented

The original `setUp()` blocks each spelled out `setup=True, default_config=True, no_output=True`. Now those defaults live only in `_make_loaded_fixture`. Anyone reading a single migrated test class no longer sees what flags were used to load the case. A one-line comment in `conftest.py` near the kwargs would help future readers grep for "why is `no_output` set on this `ss`."

---

Nothing rises to `bug` / `api-break`, so no inline comments per the spec. Phase 1 is a clean win — collection time and per-class load amortization both confirmed locally.

— independent reviewer agent

leocuisine:
Thanks for the review — addressed all four in `456108a2`. Per-finding disposition:

**#1 [correctness-risk] — applied (YAGNI route).** Dropped the public `*_loaded` fixtures entirely. Replaced with a module-level `_LOADED_CACHE` dict consumed only by `_make_fresh_fixture`'s closure. Per-process semantics match `scope="session"` but no public name exists for a future Phase 2 consumer to grab and mutate. If a read-only consumer materializes later, it earns a purpose-built fixture with the share-rules documented at the call site, not in conftest.

**#2 [style] — applied.** Factory sets `__name__` and `__doc__` on the returned fixture; `pytest --fixtures` now shows the bound name (e.g. `pjm5bus_json -- tests/conftest.py:65`) instead of the generic `_loaded` / `_fresh`.

**#3 [nit] — applied.** Collapsed numpy + pytest into a single third-party import block in `test_routine.py`.

**#4 [nit] — applied.** Hoisted load kwargs into a single `_LOAD_KWARGS` dict at the top of conftest with an inline comment explaining each (`setup`, `default_config`, `no_output`). One source of truth for "why these flags."

Suite: 348 passed, 52.50 s. flake8 clean. Re-pushing for CI.

PR #251, opened 2026-05-02
test: consolidate Family-A routine scenarios (Phase 2a)

leocuisine:
## Summary

Phase 2a of the tests refactorization. Collapses 8 single-period DC LP routine variant classes (DCOPF, DCOPF2, RTED, RTEDDG, RTEDES, RTED2, RTED2DG, RTED2ES) into one parametrized scenario suite at \`tests/routines/test_scenarios_lp_singleperiod.py\`.

Background: each of those 8 classes had byte-identical \`init\` / \`trip_gen\` / \`trip_line\` / \`set_load\` / \`vBus\` / \`dc2ac\` bodies modulo routine name. The new module parametrizes a single body and uses a small \`_ROUTINES\` table for the per-routine deltas:

- \`needs_ptdf\` for the \`*2\` family (build PTDF in fixture)
- \`has_aBus\` for the \`*2\` family (assert aBus too in vBus scenario)
- \`solver='SCIP'\` for MISOCP storage variants (RTEDES, RTED2ES), gated by \`_HAS_MISOCP\` skip
- \`_HAS_PYPOWER\` skip on \`test_dc2ac\`

Routine-unique extras (\`test_export_csv\`, \`test_align_*\`, \`test_pb_formula\`, \`test_ch_decision\`) stay in their original files. Empty classes (TestRTED, TestRTEDDG after consolidation) are removed.

Trailing \`set(value=1)\` / \`Line.alter\` resets dropped — Phase 1's per-test deepcopy fixture provides isolation.

## Numbers

| | Before | After |
|---|---|---|
| Suite tests | 348 | 348 (48 deleted methods → 48 parametrized ids) |
| Suite wall | 52.27 s | 53.49 s (+1.2 s noise) |
| Family-A LOC | 790 (across 4 files) | 199 (across 4 files + 1 new module) |
| Net diff | — | +24 / −790 |

Wall time is intentionally not the headline. The Family-A solves still happen — what changes is structural: a new Family-A variant is now a 1-line addition to \`_ROUTINES\` instead of a new class, and a failure in \`test_init[RTED2DG]\` now points unambiguously at \`ams/routines/rted.py::RTED2DG\` rather than being mixed into a per-class file.

## Phasing

This is **Phase 2a** of a planned **2-PR Phase 2**:
- 2a (this PR): Family A only — provably byte-identical bodies, lowest risk.
- 2b (follow-up after 2a merges): Families B (ACOPF on case14), C (power-flow / pypower-backend), D (multiperiod LP — ED*, ED2*), E (multiperiod MILP — UC*, UC2*).

Plan and family stratification are in \`.claude/projects/tests_refactorization/plan.md\` (private repo).

## Test plan
- [x] \`pytest tests/ -q --durations=20\` — 348 passed in 53.49 s
- [x] \`pytest tests/routines/test_scenarios_lp_singleperiod.py -q\` — 48 passed in 5.59 s
- [x] \`flake8 tests/routines/ tests/test_rtn_dcopf.py tests/test_rtn_dcopf2.py tests/test_rtn_rted.py tests/test_rtn_rted2.py\` — clean
- [ ] CI green on all 9 matrix legs
- [ ] Reviewer pass


leocuisine (commented):
## Overall verdict

The consolidation is structurally sound and does what it claims: 8 near-identical scenario classes collapse cleanly into one parametrized module backed by a `_ROUTINES` spec table, and the trimmed legacy files now contain only the variant-specific scenarios that don't fit Family A's body (`test_export_csv`, `test_align_*`, `test_pb_formula`, `test_ch_decision`). Reliance on the function-scoped deepcopy fixture from `tests/conftest.py` is correct, so dropping the trailing `set(value=1)` / `Line.alter(value=1)` resets is genuinely safe — the cached load is only ever returned via `copy.deepcopy`. However, two scenario bodies have silently changed semantics versus develop, and one of those is for the MISOCP storage variants the reviewer asked us to spot-check. (Filed as a comment review since the PR author and reviewer git user are the same; the `[correctness-risk]` items below should be treated as change-requests in spirit.)

---

### `[correctness-risk]` `test_set_load` value drifted from 0.05 to 0.1 for `RTEDES` / `RTED2ES`

In `tests/routines/test_scenarios_lp_singleperiod.py` the consolidated body uses

    ss.PQ.set(src='p0', attr='v', idx='PQ_1', value=0.1)

uniformly. On `develop`, both `TestRTEDES.test_set_load` (`tests/test_rtn_rted.py`) and `TestRTED2ES.test_set_load` (`tests/test_rtn_rted2.py`) used `value=0.05`, while only the LP variants used 0.1. The original author chose 0.05 deliberately for the storage variants — presumably because at 0.1 the strict-decrease assertion `pgs_pqt < pgs` may not hold once ESD1 charging absorbs the freed-up generation. Two acceptable fixes:

- Add a `set_load_pq1` field to `_RoutineSpec` (defaulting to 0.1, overriding to 0.05 for `RTEDES` / `RTED2ES`), then use `spec.set_load_pq1` in the test body.
- Or empirically verify 0.1 still strictly decreases `pg.v.sum()` under SCIP for both storage routines and document why the prior 0.05 was overcautious.

Right now the PR silently changes the assertion's input and the diff doesn't call it out.

### `[correctness-risk]` `test_init` for `RTEDES` / `RTED2ES` is now skipped on systems without a MISOCP solver

The `ss` fixture skips at the *fixture* level if `spec.solver == 'SCIP' and not _HAS_MISOCP`:

    if spec.solver == 'SCIP' and not _HAS_MISOCP:
        pytest.skip("No MISOCP solver is available.")

That gate is too coarse. On `develop`, `TestRTEDES.test_init` and `TestRTED2ES.test_init` carry **no** `@skip_unittest_without_MISOCP` decorator — only `init()` is called, no solver is invoked, so they run regardless of solver availability. After this PR, CI on a MISOCP-less environment will skip those two `test_init` cases that previously ran. Either:

- Move the skip to the individual tests that actually call `rtn.run(solver='SCIP')` (i.e. everything except `test_init`), or
- Have the fixture skip conditionally on a request-side flag, or
- Keep `_HAS_MISOCP` only as a marker on solver-using tests via `pytest.mark.skipif` per-test (mirroring how `_HAS_PYPOWER` already gates only `test_dc2ac`).

### `[style]` `request.node.callspec.params["ss"]` is the long way round

`request.node.callspec.params[...]` works (verified locally with pytest 9), but it's an internal-ish API the pytest docs don't bless. Cleaner options, in rough order of preference:

1. Have the `ss` fixture return both the system and the spec/routine_id:

       @pytest.fixture
       def ss(pjm5bus_json, request):
           routine_id = request.param
           spec = _ROUTINES[routine_id]
           ...
           return SimpleNamespace(system=ss, routine_id=routine_id, spec=spec)

   or just `return ss, routine_id, spec`. Tests then unpack — no `request` needed, no `_resolve`.

2. Hoist the duplicated parametrize decorator (currently repeated 7x verbatim) to a module-level alias:

       _PARAMETRIZE = pytest.mark.parametrize("ss", _ROUTINE_IDS, indirect=True, ids=_ROUTINE_IDS)

   and apply `@_PARAMETRIZE` per test.

### `[style]` `_HAS_MISOCP` duplicates `ams.shared.misocp_solvers`

`ams/shared.py` already defines `misocp_solvers = ['MOSEK', 'CPLEX', 'GUROBI', 'XPRESS', 'SCIP']` and `installed_solvers = cp.installed_solvers()`. The new module redefines both:

    _MISOCP_SOLVERS = {'MOSEK', 'CPLEX', 'GUROBI', 'XPRESS', 'SCIP'}
    _HAS_MISOCP = bool(_MISOCP_SOLVERS & set(cp.installed_solvers()))

If `ams.shared.misocp_solvers` ever changes (e.g. dropping SCIP because of an MIP-solver split), this private copy will silently desync. Please import from `ams.shared` instead:

    from ams.shared import misocp_solvers, installed_solvers
    _HAS_MISOCP = any(s in misocp_solvers for s in installed_solvers)

Same applies to `_HAS_PYPOWER`: `ams.shared.skip_unittest_without_PYPOWER` already encapsulates the import-or-skip check; you can lift its body or expose a small `has_pypower()` helper instead of duplicating the try/import. (This isn't a blocker but matters for the stated goal of "detecting the same set of solvers/installs as the original decorators".)

### `[style]` `_RoutineSpec` is a natural place for the diverging fields

Currently the table carries `needs_ptdf`, `has_aBus`, `solver`. Consider also folding in:

- `set_load_pq1` (resolves the correctness-risk above).
- `requires_solver` (bool) — used to drive whether the fixture skips on MISOCP availability for the *tests that actually run a solver*; `test_init` would set `requires_solver=False` at the test level.

This trades scattered conditionals for a slightly fatter dataclass — a good trade once the table is the single source of truth.

### `[nit]` Naming and placement

- `tests/routines/` is fine and aligns with the future Family B plan in the docstring. No objection.
- `test_scenarios_lp_singleperiod.py` reads slightly awkwardly. `test_lp_singleperiod_scenarios.py` (or `test_lp_single_period.py`) is more natural English and groups better when other Family-A modules land. Not blocking.

### `[nit]` PEP-8 import grouping

The new module mixes stdlib (`dataclasses`), third-party (`numpy`, `pytest`, `cvxpy`), and the conditional `pypower` import in an unusual order:

    from dataclasses import dataclass

    import numpy as np
    import pytest

    try:
        import pypower  # noqa: F401
        _HAS_PYPOWER = True
    except ImportError:
        _HAS_PYPOWER = False

    import cvxpy as cp

The unconditional `import cvxpy as cp` should sit with the other unconditional third-party imports above the `try/except`, not below it. Minor.

### `[nit]` Trimmed `test_rtn_dcopf2.py` redundant `build_ptdf`

Pre-existing, but `test_align_dcopf` calls `self.ss.mats.build_ptdf()` again at line 29 even though `setUp` already does so at line 22. Worth cleaning up while you're in the area.

### `[nit]` Empty `tests/routines/__init__.py`

OK as-is — matches the repo convention seen in `tests/__init__.py` and `tests/codegen/__init__.py`. Just calling out.

---

### Things I checked and that are fine

- The function-scoped deepcopy fixture (`tests/conftest.py::pjm5bus_json`, via `_make_fresh_fixture`) does what the PR claims: `_LOADED_CACHE` holds one raw load per process, and every fixture call returns `copy.deepcopy(...)`. Dropping `StaticGen.set(u=1)` and `Line.alter(u=1)` resets is therefore harmless — no leakage even within a single test body, since each test gets a fresh copy. Nothing in the consolidated bodies re-uses `ss` across functions.
- The `dc2ac` body is faithfully consolidated (assertion shape and `decimal=3` tolerance match develop).
- `test_trip_gen` / `test_trip_line` / `test_vBus` bodies match develop semantics aside from the `assertAlmostEqual(..., places=6)` → `abs(...) < 1e-6` rewrite, which is equivalent.
- `_MISOCP_SOLVERS` set matches `ams.shared.misocp_solvers` exactly (5 solvers); detection logic is equivalent modulo the duplication noted above.
- `numpy` import correctly removed from `test_rtn_dcopf.py` (no longer used) and correctly retained in the other three trimmed files (still used by remaining bodies).
- The trimmed legacy files have no dead imports, no orphaned helper code, and no broken references.

Once the two `[correctness-risk]` items are addressed I'm happy to re-review.

leocuisine:
Thanks for the thorough pass — addressed in `e17eb46f` + `fbe9e235`.

### `[correctness-risk]` `test_set_load` value drift — **fixed**
Added `set_load_pq1: float = 0.1` to `_RoutineSpec`; `RTEDES` and `RTED2ES` set it to `0.05`. The body now reads `ctx.spec.set_load_pq1`. Confirmed develop used `0.05` for both ES variants via `git show origin/develop:tests/test_rtn_rted{,2}.py`. Strict-decrease assertion holds.

### `[correctness-risk]` `test_init` skipping under no-MISOCP — **fixed**
Moved the MISOCP gate out of the fixture into a per-test helper `_skip_if_solver_missing(spec)`. Called only by tests that actually invoke `rtn.run(...)`: `test_trip_gen`, `test_trip_line`, `test_set_load`, `test_vBus`, `test_dc2ac`. `test_init` no longer skips for `RTEDES` / `RTED2ES` on solver-less CI — matches develop, where `init()` carried no decorator.

### `[style]` Long-way-round `request.node.callspec.params["ss"]` — **fixed**
Fixture now returns a frozen `_Ctx(ss, routine_id, spec, rtn)` dataclass. Tests take `ctx` and unpack — no more `request` introspection, `_resolve` helper deleted. Also hoisted the duplicated parametrize decorator to module-level alias `_PARAMETRIZE_ROUTINES`.

### `[style]` `_HAS_MISOCP` desync risk — **fixed**
Now `from ams.shared import installed_solvers, misocp_solvers`; `_HAS_MISOCP = bool(set(misocp_solvers) & set(installed_solvers))`. No more local copy. Left `_HAS_PYPOWER` as a local try-import — `ams.shared` only exposes the unittest decorator, not a plain bool, and the try/import is two lines so the abstraction would cost more than it saves.

### `[style]` `_RoutineSpec` should grow fields — **partially adopted**
Added `set_load_pq1` (resolves the correctness-risk above). Skipped `requires_solver` since the per-test `_skip_if_solver_missing(spec)` call already serves that purpose more clearly than a flag the fixture would have to special-case.

### `[nit]` Module name `test_scenarios_lp_singleperiod.py` — **kept as-is for now**
Valid grammar concern, but renaming now also requires renaming the symmetric Family-D / Family-E modules I'll add in PR 2b. I'd rather pick the final name in 2b once the full set is visible. Filed as a thread to revisit.

### `[nit]` PEP-8 import grouping — **fixed**
`import cvxpy as cp` is gone (replaced by `ams.shared` import); the `try: import pypower` block now sits below the unconditional third-party imports, matching the standard order.

### `[nit]` Redundant `build_ptdf` in `test_rtn_dcopf2.py::test_align_dcopf` — **fixed**
Dropped in `fbe9e235`. setUp already builds it.

### `[nit]` Empty `tests/routines/__init__.py` — kept (matches repo convention).

Local: 48 passed in 5.43 s on the new module; full suite 348 passed (re-running on push). flake8 clean.

PR #252, opened 2026-05-02
test: consolidate Family B-E routine scenarios (Phase 2b)

leocuisine:
## Summary

Phase 2b of the test refactorization. Consolidates Families B–E (per-routine scenario duplication across six legacy test files) into three parametrized scenario modules under `tests/routines/`, mirroring the Phase 2a pattern (PR #251).

| Family | Routines | New module |
|---|---|---|
| C — power-flow | DCPF, PFlow (case14); DCPF1, PFlow1, DCOPF1, ACOPF1 (pjm5bus_demo) | `test_scenarios_pflow.py` |
| D — multiperiod LP | ED, EDDG, EDES, ED2, ED2DG, ED2ES | `test_scenarios_lp_multiperiod.py` |
| E — multiperiod MILP | UC, UCDG, UCES, UC2, UC2DG, UC2ES | `test_scenarios_milp_multiperiod.py` |

Per-routine differences are captured in a frozen `_RoutineSpec` dataclass per family — case fixture, generator id, solver, MISOCP/PYPOWER gates, alignment reference, etc. The 2nd-gen `align_*` and `pb_formula` tests live in their own parametrized blocks targeting only routines that have an `align_ref`.

## Numbers

- **Test count preserved**: 348 collected, same as post-Phase-2a baseline. 151 of those are now parametrized routine scenarios (48 from Phase 2a + 103 added here).
- **Wall time**: 348 passed in **58.55 s** vs. 53.49 s post-Phase-2a baseline (+5 s noise; well under the 61.8 s pre-refactor baseline).
- **Net diff**: +582 / -1889 LOC.

## Routines intentionally untouched

- **ACOPF** (Family B) — single routine, no consolidation value.
- **OPF** (gurobi_optimods, dual DC/AC mode) — doesn't fit a single-shape body; left for a later one-off cleanup.

## Surprise during implementation

`UC2ES.test_align_uces` on develop only compares `obj` and `ugd` to UCES; the `pg`/`aBus`/`plf` comparisons are commented out (generation allocation diverges while the objective matches — see source for context). Preserved verbatim via an `align_full=False` flag on the spec.

## Test plan

- [x] `pytest tests/routines/ -q` — 151 passed in 34.22 s
- [x] `pytest tests/ -q --durations=15` — 348 passed in 58.55 s
- [x] `flake8 tests/routines/` clean
- [ ] CI green on push
- [ ] Independent reviewer agent — spawned on PR-open


leocuisine (commented):
Reviewed PR #252 against the deleted legacy modules on `develop`. The consolidation pattern is solid and the test-count math works out (25 + 36 + 42 = 103 across Families C/D/E). One real value drift, plus a few smaller items. Findings below.

### 1. `[bug]` ED2ES `loc_offtime` drift — test_scenarios_lp_multiperiod.py:96

```python
loc_offtime = np.array([0, 2, 4]) if ctx.spec.solver == 'CLARABEL' else np.array([0, 2])
```

The branch keys on `solver`, lumping ED2ES (SCIP) with EDES (SCIP). But on `develop`:

| Routine        | Solver   | `loc_offtime`     |
|----------------|----------|-------------------|
| ED, EDDG       | CLARABEL | `[0, 2, 4]`       |
| ED2, ED2DG     | CLARABEL | `[0, 2, 4]`       |
| **EDES**       | SCIP     | **`[0, 2]`**      |
| **ED2ES**      | SCIP     | **`[0, 2, 4]`**   |

Source: `tests/test_rtn_ed2.py::TestED2ES::test_trip_gen` on develop uses `loc_offtime = np.array([0, 2, 4])`. The new code shrinks ED2ES coverage from 3 off-time slots to 2 — exactly the kind of silent value drift Phase 2a missed for `set_load`. The keying needs to be per-routine-id (e.g., `EDES → [0, 2]`, everything else `[0, 2, 4]`), not per-solver.

### 2. `[correctness-risk]` `test_align` run order swapped for UC2 — test_scenarios_milp_multiperiod.py:158-159

In `develop`'s `tests/test_rtn_uc2.py::TestUC2::test_align_uc`, the order is `UC.run(...)` first, then `UC2.run(...)`. UC2DG/UC2ES (and all ED2*) call the 2nd-gen first. The new module unconditionally runs `ctx.rtn` (2nd-gen) first, then `ref` (1st-gen). For UC2 specifically this reverses the original ordering. Both routines re-build their own CVXPY problem on `.run()`, but they share `ss.StaticGen.u` mutated by `_initial_guess()` (called once on `ctx.rtn` in the fixture). Likely benign for SCIP MILP determinism, but it is a behaviour change vs. develop and worth documenting or matching exactly.

### 3. `[test-gap]` `test_init` runs after `_initial_guess()` in milp module — test_scenarios_milp_multiperiod.py:65-72, 94-97

The fixture invokes `rtn._initial_guess()` for every test, including `test_init`. On develop, `setUp` does the same, so the legacy `test_init` also ran on a system where ~30% of generators were forced offline. This appears intentional and `_initial_guess()` is idempotent (uc.py:273-275 short-circuits if any gen is already off), so `init()`'s internal `_initial_guess()` call is a no-op. Not a blocker, but also not strictly the same: legacy `test_init` re-validated initialization on the modified system; the new test inherits that behaviour by accident-of-fixture rather than by intent. A one-line comment in the fixture would help future readers.

### 4. `[style]` Duplicate `_HAS_MISOCP` / `_HAS_PYPOWER` constants

Defined in `tests/routines/test_scenarios_lp_multiperiod.py:9`, `tests/routines/test_scenarios_milp_multiperiod.py:30`, and `tests/routines/test_scenarios_pflow.py:33`. Phase 2a's `tests/routines/test_scenarios_lp_singleperiod.py` likely has the same constant. Hoist to `tests/conftest.py` (or a small `tests/routines/_solver_gates.py`) to keep the gate definition canonical.

### 5. `[nit]` Empty `np.zeros_like(loc_offtime)` reference uses int dtype — test_scenarios_lp_multiperiod.py:103

`loc_offtime` is `np.array([0, 2, 4])` (int). `np.zeros_like(loc_offtime)` is int-zero. The `assert_almost_equal(int_zeros, float_pg, decimal=6)` works in NumPy but is a latent type oddity inherited verbatim from develop. Mentioning only because the consolidation is a fine moment to use `np.zeros(len(loc_offtime))` instead. Skip if you'd rather keep parity.

### 6. `[nit]` ED2ES `align_decimals=3` — test_scenarios_lp_multiperiod.py:36

Confirmed match with develop (`tests/test_rtn_ed2.py::TestED2ES::test_align_edes` uses `DECIMALS = 3`). Just calling out that this is preserved correctly.

---

**Summary**: 1 bug (#1), 1 correctness-risk (#2), 1 test-gap (#3), 2 style/nit (#4, #5). Only #1 is a blocker — please pin `loc_offtime` per routine-id rather than per solver before merge. The rest can ship as-is or with a follow-up.

tests/routines/test_scenarios_lp_multiperiod.py
```diff
@@ -0,0 +1,207 @@
...
+
+@_PARAMETRIZE_ROUTINES
+def test_trip_gen(ctx):
+    """Verify EDTSlot.ug takes effect, and StaticGen.u is ignored."""
+    _skip_if_solver_missing(ctx.spec)
+
+    # a) EDTSlot.ug.v drives per-slot generator status
+    stg = 'PV_1'
+    stg_uid = ctx.ss.StaticGen.get_all_idxes().index(stg)
+    loc_offtime = np.array([0, 2, 4]) if ctx.spec.solver == 'CLARABEL' else np.array([0, 2])
```
leocuisine: [bug] Value drift vs develop: `loc_offtime` keys on `solver`, but on develop EDES uses `[0, 2]` and ED2ES uses `[0, 2, 4]` — both use SCIP, so this branch shrinks ED2ES coverage. Pin per routine-id (e.g. `EDES → [0, 2]`, everything else `[0, 2, 4]`).

leocuisine (commented):
Thanks for the careful review. Pushed `8ec8ee23` addressing all four actionable findings:

1. **Bug — `loc_offtime` per-routine** *(blocker, fixed)*
   You were right — keying on solver lumped ED2ES with EDES. Confirmed develop has `[0, 2, 4]` for ED/EDDG/ED2/ED2DG/ED2ES and `[0, 2]` only for EDES. Added a `loc_offtime` field on `_RoutineSpec` defaulting to `(0, 2, 4)` and overridden to `(0, 2)` only for EDES.

2. **Correctness-risk — UC2 `test_align` run order** *(fixed)*
   Confirmed: legacy `TestUC2.test_align_uc` runs `UC.run(...)` before `UC2.run(...)`, while UC2DG/UC2ES run the 2nd-gen first. Added an `align_ref_first` flag (True only for UC2) so the new parametrized body preserves the legacy ordering verbatim. SCIP is deterministic on these problems so it would likely have been benign, but matching develop avoids any silent behavior change.

3. **Test-gap — `_initial_guess()` in fixture** *(comment added)*
   `_initial_guess()` is idempotent and solver-free, so the legacy per-method `setUp` invocation and the new per-fixture invocation are semantically equivalent. Added a clarifying comment in the fixture.

4. **Style — duplicate `HAS_MISOCP` / `HAS_PYPOWER`** *(fixed)*
   Hoisted both into `tests/conftest.py` as a single source of truth. All three new modules + the Phase 2a `test_scenarios_lp_singleperiod.py` now import from there.

Skipped the int-zero-vs-float nit — `np.testing.assert_almost_equal(np.zeros_like(int_array), pg_pv1[loc_offtime], decimal=6)` is what the legacy bodies did, and the upcast is fine here.

Full suite: 348 passed in 56.73 s post-fix-up.

PR #253, opened 2026-05-02
test: parametrize known-good MATPOWER cross-checks (Phase 3)

leocuisine:
## Summary

Phase 3 of the test refactorization. Collapses `tests/test_known_good.py`'s three case-specific unittest classes (`TestKnownResults` / `TestKnownResultsIEEE39` / `TestKnownResultsIEEE118`) into one parametrized scenario module under `tests/routines/`. Same 4 routines (DCPF, PFlow, DCOPF, Matrices) × 3 cases (case14, case39, case118) = 12 ids, now driven by a frozen `_CaseSpec` dataclass.

## Per-case differences captured in spec

- **`dcpf_pg_sum`**: case14 keeps element-wise pg coverage; case39/case118 only assert the total (legacy bodies did the same).
- **`dcpf_normalize_aBus`**: only case118 normalizes by subtracting the slack-bus angle before comparing — MATPOWER and AMS picked different reference angles on that case.
- **`dcopf_places`**: tighter (4) for case14, looser (2) for the larger systems.
- **`matrices_lodf_atol`**: `1e-2` for case14, `10` for case39/case118.

All tolerances + shapes preserved verbatim from the legacy bodies.

## conftest.py upgrade

- `benchmark_mpres` fixture: loads `matpower/benchmark.json` once per worker process (was re-read per-class on develop).
- `case39` and `case118` deepcopy fixtures added alongside the existing `case14`.

## Numbers

- 348 collected tests, **no count change**.
- 348 passed in **60.04 s** vs. 56.73 s post-Phase-2b baseline (+3 s noise).
- New module on its own: 12 passed in 2.03 s.
- Net diff: **+185 / -267 LOC**.

## Test plan

- [x] `pytest tests/routines/test_scenarios_known_good.py -v` — 12 passed in 2.03 s
- [x] `pytest tests/ -q` — 348 passed in 60.04 s
- [ ] CI green
- [ ] Independent reviewer agent — spawned on PR-open


leocuisine (commented):
**Independent reviewer pass — PR #253 (Phase 3: parametrize known-good MATPOWER cross-checks)**

Read the diff and ran `pytest tests/routines/test_scenarios_known_good.py -v` against `master` head with the PR applied: **12 passed in 2.63 s**, IDs match `{test_dcpf,test_pflow,test_dcopf,test_matrices} × {case14,case39,case118}` — same 12 names the legacy file produced. Diffed each branch against `origin/develop:tests/test_known_good.py` line-by-line. The refactor is faithful.

### Verifications

1. **`assertAlmostEqual` ↔ `round(diff, places) == 0` equivalence** — confirmed. CPython's `unittest.TestCase.assertAlmostEqual(a, b, places)` runs `round(abs(a-b), places) == 0` (verified via `inspect.getsource`). The new code uses `round(diff, places) == 0` (no `abs`). Because Python's banker's rounding is symmetric around zero (`round(-x, n) == -round(x, n)`), `round(diff, n) == 0  ⇔  round(abs(diff), n) == 0`. Equivalent. `dcopf_places` values carried over verbatim: case14=4, case39=2, case118=2. (`tests/routines/test_scenarios_known_good.py:144-146`)
2. **`dcpf_pg_sum` shape preserved** — case14 spec `dcpf_pg_sum=False` → element-wise pg compare (matches legacy `TestKnownResults.test_DCPF_case14`); case39/case118 spec `dcpf_pg_sum=True` → `pg.sum()` compare (matches legacy 39/118 bodies). (`test_scenarios_known_good.py:46-54,110-116`)
3. **`dcpf_normalize_aBus` for case118** — algebraically equivalent. Legacy: LHS `(aBus_v - aBus_v[0]) * rad2deg`, RHS `aBus_mp - aBus_mp[0]`. New: `aBus_sp = aBus_v * rad2deg; aBus_sp -= aBus_sp[0]; aBus_mp -= aBus_mp[0]`. Distributing `rad2deg` over the subtraction is exact (scalar multiply). The `dcpf_normalize_aBus=False` path on case14/case39 also matches legacy (no normalization). (`test_scenarios_known_good.py:103-108`)
4. **`_load_ptdf` / `_load_lodf`** — identical to develop's `load_ptdf` / `load_lodf` apart from taking `mpres_case` directly instead of `(mpres, case)`; same `_NaN_→0` for PTDF, same `_NaN_/-_Inf_/_Inf_→nan` + `fill_diagonal(-1)` for LODF. The `ptdf[np.isnan(ptdf_mp)] = nan` line on `test_scenarios_known_good.py:163` is a no-op (PTDF NaNs are pre-substituted to 0) — but this is also a no-op on develop, so no behavior change. Worth a one-line comment in a follow-up but not in scope here. (`test_scenarios_known_good.py:79-95`)
5. **Tolerance table** — `rtol/atol=1e-2/1e-2` everywhere on develop preserved verbatim; per-case `matrices_lodf_atol` (1e-2 / 10 / 10) preserved verbatim. (`test_scenarios_known_good.py:46-54,166-170`)
6. **Imports** — `from ams.shared import nan` IS used (lines 90, 163, 164). No dead imports. `import json` is intentionally function-local in `conftest.py:105` to keep top-level import light; harmless.

### Findings

1. **[nit] `tests/conftest.py:104`** — `benchmark_mpres` docstring says *"Lazy session-cache of `matpower/benchmark.json`"* but the fixture is function-scoped (default). Behavior is correct because `_BENCHMARK_CACHE` is module-level (so it's effectively per-process / "session-cached"), but the wording can mislead a reader who expects `scope="session"`. Suggest tweaking to e.g. *"Function-scoped fixture returning a per-process cached load of `matpower/benchmark.json`."*
2. **[nit / future-proofing] `tests/conftest.py:99-109`** — `_BENCHMARK_CACHE['mpres']` is returned by reference; a future test that mutates `mpres_case` would poison every subsequent test. Today's tests don't mutate, so this is **not a current bug** and I would not block on it. If you want belt-and-braces, either (a) `copy.deepcopy` the dict on return (cheap relative to test wall-time) or (b) leave a `# CAUTION: shared mutable — read-only contract` comment. Mention only because the docstring wording (#1) might be read as "you get a fresh object each time."
3. **[nit] `tests/routines/test_scenarios_known_good.py:163`** — the `ptdf[np.isnan(ptdf_mp)] = nan` line is a no-op (`_load_ptdf` substitutes `_NaN_` → 0, never NaN). Inherited from develop's `load_ptdf`. Not introduced by this PR; flagging only because it would be a natural cleanup target in a future pass.

### Summary

| Class | Count |
|-------|-------|
| bug | 0 |
| correctness-risk | 0 |
| api-break | 0 |
| test-gap | 0 |
| docs | 0 |
| style | 0 |
| nit | 3 |

**No blockers.** Phase 3 is a clean, faithful collapse of three near-duplicate unittest classes into one parametrized scenario module — tolerances, shapes, and per-case quirks (case14 element-wise pg, case118 slack-bus normalization, looser DCOPF/LODF on the larger systems) are all preserved verbatim. LGTM.

leocuisine (commented):
Thanks — agreed on all three nits, none worth a separate commit:

1. **`benchmark_mpres` docstring "session-cache"** — fair, the fixture is function-scoped with module-level caching. Will tidy the wording in the next conftest-touching PR.
2. **`_BENCHMARK_CACHE['mpres']` shared-mutable risk** — theoretical; tests treat `mpres` as read-only. Worth keeping in mind if a future test ever wants to mutate the dict.
3. **`ptdf[np.isnan(ptdf_mp)] = nan` no-op** — inherited from develop verbatim. Removing it would be a behavior-equivalent cleanup but out of scope for the consolidation PR.

Will land as-is once CI is green.

PR #248, opened 2026-05-02
docs: cvxpy-namespace migration guide (Step 6)

leocuisine:
## Summary

Closes the cvxpy-namespace passthrough project (Step 6 of 6).

- New page **`docs/source/modeling/migration_cvxpy_namespace.rst`** — dedicated migration guide for the v1.2.2 `e_str` authoring contract change. Pulls before/after from PR #244's actual `rted.py` diff (SFR balance, reserve source, quadratic cost, ESD1 SOC), explains the rationale (`dot` was a numpy false-friend, `mul` was 3-char inertia), and covers the full surface: substitution table, customization patterns, reserved-name guard, `formulation_source 'sub_map' -> 'eval'` rename, `SymProcessor.sub_map`/`val_map` removal, Constraint relational-operator dispatch.
- **`addConstrs` docstring** (`ams/routines/routine.py:1240`) gains a canonical-CVXPY blurb + two short examples (`pg - pmax` and the new `pg <= pmax` form) so anyone landing from autosummary sees the right pattern.
- **`docs/source/modeling/example.rst`** — orphaned "Example - Customize Formulation" pointer now resolves to `ex8.ipynb` under `:ref:scripting_examples`, with a cross-link to the migration guide.
- **`docs/Makefile`** — new `make distclean` target additionally wipes the gitignored autosummary + nbsphinx-copied notebook trees (`source/_generated`, `source/modeling/_generated`, `source/_examples`). `make clean` is unchanged (still build-only / fast). Bundled here for convenience.

## Test plan

- [x] `make html SPHINXOPTS='-W --keep-going -j auto'` builds clean (no warnings).
- [x] New page renders at `build/html/modeling/migration_cvxpy_namespace.html`.
- [x] All four `:ref:` cross-links (`migration_cvxpy_namespace`, `routine`, `scripting_examples`, `ReleaseNotes`) resolve.
- [x] `make clean` and `make distclean` both verified locally — `clean` leaves the generated trees alone, `distclean` wipes them.
- [ ] Reviewer agent — independent reviewer reads the diff for accuracy of the worked examples vs the actual PR #244 commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #256, opened 2026-05-02
opt: tighten eval-fallback with LHS-zero check + DCP visibility

leocuisine:
## Summary

Three small follow-ups to the cvxpy-namespace passthrough (#244-#248) and `is_eq` retirement (#249), all surfacing user-customization mistakes that previously failed silently or far from the cause. Architecture-review items L5 and M4 from the 2026-05-02 audit.

- **L5** — `Constraint.parse` now validates that `e_str` ends with `<= 0` / `== 0` / `>= 0`. A constraint authored as `pg <= pmax` solves correctly but produces a numpy bool array as `OptzBase.e` (instead of the LHS slack), because `eval_e_str_numeric`'s trailing-op stripper only matches `<op> 0`. The new `assert_constraint_lhs_zero` guard in `ams/opt/_runtime_eval.py` surfaces the mismatch immediately with an actionable error.
- **M4a** — `eval_e_str` now logs a per-item `WARNING` when the result reports `is_dcp() is False`. The codegen path skips eval entirely, so this only fires on user customizations / `addConstrs` / mixin overrides — exactly where silent DCP drift is most likely after the namespace passthrough opened the full `cp.*` surface.
- **M4b** — `RESERVED_CVXPY_ATOM_NAMES` is auto-derived from `cvxpy.atoms` at import time, unioned with the static fallback so the guarded set can only grow. Modern CVXPY exposes ~116 atom callables vs the static set of 19; future atoms are guarded automatically.
- **M4c** — Updated the trust-boundary comment in `eval_e_str` to reflect that user code rides the same path post-passthrough; the prior framing understated the actual reach.

## Test plan

- [x] `pytest tests/opt/test_runtime_eval.py` (15 new cases) — pass
- [x] `pytest tests/opt/ tests/codegen/ tests/cli/` — 42 pass
- [x] `pytest tests/routines/ tests/core/` — 244 pass
- [x] `pytest tests/` (full suite) — 363 pass, 0 fail
- [x] Spot-checked DCOPF solve on `case5.m` against pre-change output (unchanged)

### DIFF
--- a/ams/opt/_runtime_eval.py
+++ b/ams/opt/_runtime_eval.py
@@ -157,8 +157,11 @@ def assert_constraint_lhs_zero(item, e_str):
             f"with `<= 0`, `== 0`, or `>= 0` (LHS-zero authoring shape). "
             f"Got: {e_str!r}. Move every term to the LHS — "
             f"e.g. `pg - pmax <= 0` instead of `pg <= pmax`. "
-            f"Without the trailing-zero shape, `OptzBase.e` returns a "
-            f"numpy bool array instead of the LHS slack."
+            f"The trailing zero is matched literally (`0`, not `0.0` "
+            f"or `(0)`) so the same regex strips it cleanly inside "
+            f":func:`eval_e_str_numeric`. Without the trailing-zero "
+            f"shape, `OptzBase.e` returns a numpy bool array instead "
+            f"of the LHS slack."
         )
 
 
--- a/ams/opt/_runtime_eval.py
+++ b/ams/opt/_runtime_eval.py
@@ -88,12 +88,18 @@ def eval_e_str(item, e_str):
         'cp': cp, 'np': np, 'sps': sps,
         'r': RoutineNS(rtn), 'self': item,
     }
-    # Trust boundary: e_str is repo-authored in ams/routines/ or
-    # comes from trusted downstream code (notebook authors, mixins).
-    # Same boundary as the codegen-emitted pycode — both run user-
-    # accessible source through Python. No untrusted-input ingestion.
+    # Trust boundary: e_str is Python source eval'd in a curated
+    # namespace (cp/np/sps + the routine's symbol proxy). Authors come
+    # from three places — repo source under ``ams/routines/``, mixin
+    # overrides, and downstream user code (notebook ``addConstrs``,
+    # ``item.e_str = ...`` assignments after cvxpy-namespace passthrough
+    # opened the full ``cp.*`` surface). All three ride the same path;
+    # treat them with the same trust as the codegen-emitted pycode —
+    # both run user-accessible source through Python. No untrusted-
+    # input ingestion contract here, so authors are responsible for
+    # what their ``e_str`` evaluates to.
     try:
-        return eval(code, {}, local_vars)  # nosec B307
+        result = eval(code, {}, local_vars)  # nosec B307
     except Exception as exc:
         raise Exception(
             f"Error evaluating {item.class_name} <{item.name}> via e_str.\n"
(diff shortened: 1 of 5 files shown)


leocuisine (commented):
LGTM with minor observations.

I read the PR diff against `ams/routines/`, `ams/opt/`, and `ams/prep/generator.py`, and verified the three claims (L5 LHS-zero validator, M4a DCP warning, M4b dynamic atom-name derivation) against runnable behavior.

**What I verified**

1. **Correctness — `assert_constraint_lhs_zero`** (`ams/opt/_runtime_eval.py:137-162`, `ams/opt/constraint.py:110-111`)
   Walked every `Constraint(... e_str=...)` declaration in `ams/routines/` (dcopf, dcpf, dopf, ed, rted, uc, dcopf2, pflow, pypower) including the dynamically-composed strings (`SOCb` / `SOCb0` / `tcdr` / `tddr` / `pb`). All current built-ins end with the `<op> 0` shape. The regex `r'\s*(==|<=|>=)\s*0\s*$'` correctly accepts all of them. The validator is also called on codegen-path items because `e_str` is preserved on the item — so it covers both authoring paths as the docstring claims. No built-in regression.

2. **DCP warning visibility** (`ams/opt/_runtime_eval.py:110-130`)
   Confirmed `is_dcp()` is defined on `cp.constraints.Constraint` and on every `cp.Expression` subtree, and that `cp.multiply(pg, pg) - 1 <= 0` round-trips with `is_dcp() is False`. Numeric/literal returns (e.g. the `e_str='0'` placeholder objective in `pflow.py` / `dcpf.py`) lack `is_dcp` and the `getattr(...) + callable(...)` guard skips them — no false positives. The exception swallow with `ok = True` is the right default (DCP introspection failures must not block eval).

3. **`RESERVED_CVXPY_ATOM_NAMES` derivation** (`ams/prep/generator.py:51-90`)
   On the local CVXPY 1.6.5 install the derivation yields 113 names (vs static 19). Intersected the derived set with every symbol name declared across `ams/routines/` (Vars, RParams, Services, Expressions, Constraints, config keys including `t`, `tol`, `verbose`, `pf_alg`, etc.) — **zero collisions** with current built-ins. The static-fallback union guarantees the guarded set never narrows. `cp.atoms` is not a documented public surface but has been stable for many releases, and the `callable(getattr(cp, name))` filter is the right safety belt.

4. **Tests** (`tests/opt/test_runtime_eval.py`)
   Three classes exercise each tightening directly. The `assertLogs` sentinel pattern in `test_no_warning_on_dcp` is the standard workaround for `assertLogs` requiring at least one record — comment explains. Codegen-path validator coverage is implicit (the broader suite's `case5` setup calls `init()` which parses every built-in Constraint through the validator).

**Findings**

1. `docs` — `docs/source/release-notes.rst:35` — `v1.2.2 (unreleased)` is already a shipped tag (`git tag` shows v1.2.2). This is a pre-existing stale label, not introduced by this PR, but worth a one-line follow-up to flip it to a real release date so the new `v1.2.3 (unreleased)` block isn't sitting above a misleading sibling. Suggested fix: separate doc PR re-labeling v1.2.2 with its release date.

2. `nit` — `ams/opt/_runtime_eval.py:134` — The trailing-zero regex requires the literal token `0`. `... <= 0.0`, `... <= (0)`, or `... <= 0  # comment` would be rejected. No current routine authors any of those, so no breakage today, but the error message ("Move every term to the LHS — e.g. `pg - pmax <= 0`") doesn't hint at the literal-zero strictness if someone tries `0.0`. Optional: broaden regex to `r'\s*(==|<=|>=)\s*0(\.0*)?\s*$'`, or call it out in the error.

3. `nit` — `ams/opt/_runtime_eval.py:119-130` — `is_dcp()` walks the expression tree on every eval-fallback call. Negligible for typical user customizations, but a routine that hits `eval_e_str` per-item over a large `addConstrs` loop would pay it linearly. Acceptable; flagging only because the M4a comment frames this as a cheap visibility hook and that's roughly true but not free.

4. `nit` — `tests/opt/test_runtime_eval.py:88-97` — `test_rejects_bad_user_constraint` mutates `self.ss.DCOPF.plflb.e_str = 'plf <= rate_a'` then calls `parse()` directly. Coverage is fine, but no test asserts the validator fires on a routine that hasn't been initialized yet (e.g. raw `Constraint(name='c', e_str='pg <= pmax').parse()` — would need an `om` attached, so the existing test shape is the practical one). Not a gap to fix; just noting that the validator's "covers both paths" claim rests on the pattern that init→parse runs over every constraint.

Single review, no per-line bugs to anchor at the diff.

leocuisine:
Thanks for the review. Dispositions:

1. **`docs` — stale `v1.2.2 (unreleased)` label**: agreed it's pre-existing and not for this PR. Filed as a separate cleanup; this PR keeps the new `v1.2.3 (unreleased)` block above the misleading sibling.

2. **`nit` — trailing-zero regex strictness**: kept the strict `0` literal (the same regex powers the strip in `eval_e_str_numeric`, so any broadening would have to land in both places to stay coherent). Updated the error message in `36a84af8` to call out the literal-zero requirement so a user trying `0.0` / `(0)` sees why it's rejected.

3. **`nit` — `is_dcp()` per-call cost**: acknowledged. The eval-fallback path is rarely hit at scale (codegen handles built-ins); if profiling ever shows it, the warning is easily gated behind a logger-level check.

4. **`nit` — test coverage observation on validator firing path**: agreed this is "not a gap to fix" — every routine test exercises `init()` which parses every Constraint through the validator, so the codegen path coverage is real (just indirect).

CI lint failure on the first push (3 unused imports) was cleaned up in `2b308958`.

PR #261, opened 2026-05-03
feat: report path arg + load_csv round-trip (closes #186)

### LINKED ISSUE #186: Allow input file path in ``system.report``
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

leocuisine:
## Summary

Bundles three related changes targeting **v1.2.4** on `develop`:

1. **#186 — `system.report(path=...)`** — per-call output path. Explicit `path` overrides `no_output=True` (per-call intent wins). `path` may be a directory or a full file path; resolved via the existing `get_export_path` helper.
2. **Path-default consistency** — `get_export_path` now defaults to `system.files.output_path` (when set) before falling back to CWD, so `RoutineBase.export_csv` / `export_json` agree with `Report` on where files land. Empty `output_path` (the `FileMan` default) still falls through to CWD — no test fixture churn.
3. **Round-trip load API** — new `RoutineBase.load_csv(path)` inverts `export_csv` (single-period via `'T1'` sentinel; multi-period reindexed to `self.timeslot.v`). Both `load_json` and `load_csv` now set `converged=True` / `exit_code=0` on success so post-load `system.report()` and `routine.get(...)` behave as if the routine just solved.

**Behavior change** (flagged in v1.2.4 release notes): `load_json` previously left `converged` untouched; it now sets `converged=True` on success. The "load and inspect" workflow this enables is the AMS analog of ANDES `TDS.run(from_csv=...)`.

**Docstring convention introduced** — every modified docstring carries a `Notes` block with a "Differences from ANDES:" sub-section, to keep the cognition burden low for users coming from ANDES.

## Test plan

- [x] `pytest tests/io/ -v` — 51/51 (5 new: `test_report_path_dir`, `test_report_path_file`, `test_load_csv_DCOPF`, `test_load_csv_ED`, `test_export_default_uses_output_path`)
- [x] Full suite `pytest tests/ -x -q` — **368 passed**
- [x] Manual smoke: solve DCOPF → `export_csv` → fresh system → `load_csv` → `pg` matches within 1e-5 → `report(path=...)` produces a readable report
- [x] flake8 clean on touched files

## Out of scope

- `system.load_results(path)` dispatcher (per-routine call sites are already short)
- Dedicated AMS↔ANDES divergence docs page (per-method `Notes` blocks first)
- MatProcessor `export_csv` default change (not coupled to `system.files`)

### DIFF
--- a/ams/routines/routine.py
+++ b/ams/routines/routine.py
@@ -757,7 +757,21 @@ class RoutineBase:
         bool
             True if the loading is successful, False otherwise.
 
+        Notes
+        -----
+        Differences from ANDES: ANDES has no per-routine ``load_json``;
+        the closest analog is :meth:`andes.routines.tds.TDS.run` with
+        ``from_csv=`` for time-series replay. AMS reads scheduling
+        results back into existing :class:`Var` / :class:`ExpressionCalc`
+        storage and marks the routine as converged so subsequent
+        :meth:`get` and :meth:`ams.system.System.report` calls behave
+        as if the routine just solved.
+
         .. versionadded:: 1.0.13
+
+        .. versionchanged:: 1.2.4
+           A successful load now sets ``self.converged = True`` and
+           ``self.exit_code = 0``.
         """
         try:
             with open(path, 'r') as f:
@@ -793,6 +807,113 @@ class RoutineBase:
                         exprc.v = np.array(values)
                     except Exception as e:
                         logger.warning(f"Failed to assign values to exprc '{key}': {e}")
+
+        self.converged = True
+        self.exit_code = 0
+        logger.info(f"Loaded results from {path}")
+        return True
+
+    def load_csv(self, path):
+        """
+        Load scheduling results from a csv file produced by
+        :meth:`export_csv`.
+
+        For multi-period routines the ``Time`` column is used to
+        re-align rows to ``self.timeslot.v``. For single-period
+        routines, the ``Time`` column carries the literal sentinel
+        ``"T1"`` and a single row is expected.
+
+        Parameters
+        ----------
+        path : str
+            Path of the csv file to load.
+
+        Returns
+        -------
+        bool
+            ``True`` if the loading is successful, ``False`` otherwise.
+
+        Notes
+        -----
+        Round-trip values are clipped to 6 decimals on export (see
+        :func:`gather_results`) — reload is lossy at the 1e-6 level,
+        which is fine for inspection but not for bit-exact replay.
+
+        Differences from ANDES: ANDES has no per-routine ``load_csv``;
+        the closest analog is :meth:`andes.routines.tds.TDS.run` with
+        ``from_csv=`` for time-domain replay. AMS reads scheduling
+        results back into existing :class:`Var` / :class:`ExpressionCalc`
+        storage and marks the routine as converged so subsequent
+        :meth:`get` and :meth:`ams.system.System.report` calls behave
+        as if the routine just solved.
+
+        .. versionadded:: 1.2.4
+        """
+        try:
+            df = pd.read_csv(path)
+        except Exception as e:
+            logger.error(f"Failed to load CSV file: {e}")
+            return False
+
+        if 'Time' not in df.columns:
+            logger.error(f"CSV file '{path}' missing required 'Time' column.")
+            return False
+
+        if not self.initialized:
+            self.init()
+
+        # Single-period CSVs use the literal 'T1' sentinel from
+        # ``initialize_data_dict`` (a single row); multi-period CSVs
+        # carry timeslot idx strings (e.g. 'EDT1', 'EDT2', ...).
+        time_vals = df['Time'].astype(str).values
+        single_period = len(time_vals) == 1 and time_vals[0] == 'T1'
+
+        if not single_period:
+            if not hasattr(self, 'timeslot'):
+                logger.error(
+                    f"CSV at '{path}' has multi-period rows but routine "
+                    f"<{self.class_name}> has no 'timeslot'.")
+                return False
+            slots = list(self.timeslot.v)
+            df = df.set_index('Time').reindex(slots)
+            missing = df.index[df.isna().all(axis=1)].tolist()
+            if missing:
+                logger.warning(
+                    f"CSV missing rows for timeslots {missing}; "
+                    f"loaded values for those slots will be NaN.")
+
+        for items in (self.vars, self.exprcs):
+            for key, item in items.items():
+                if item.owner is None:
+                    continue
+                idxes = item.get_all_idxes()
+                cols = [f'{key} {dev}' for dev in idxes]
+                present = [c for c in cols if c in df.columns]
+                if not present:
+                    logger.debug(f"CSV has no columns for '{key}'; skipping.")
+                    continue
+                if len(present) < len(cols):
+                    missing_devs = [d for d, c in zip(idxes, cols)
+                                    if c not in df.columns]
+                    logger.warning(
+                        f"CSV missing devices {missing_devs} for '{key}'; "
+                        f"those entries will be NaN.")
+                # Build a (n_dev,) or (n_slot, n_dev) frame, reindex to
+                # full device list to get NaN columns for absent devices.
+                sub = df[present].reindex(columns=cols)
+                try:
+                    if single_period:
+                        item.v = sub.iloc[0].to_numpy(dtype=float)
+                    else:
+                        # CSV is one row per slot, columns per device;
+                        # var.v shape is (n_dev, n_slot) → transpose.
+                        item.v = sub.to_numpy(dtype=float).T
+                except Exception as e:
+                    logger.warning(
+                        f"Failed to assign values to '{key}': {e}")
+
+        self.converged = True
+        self.exit_code = 0
         logger.info(f"Loaded results from {path}")
         return True
 
--- a/ams/routines/routine.py
+++ b/ams/routines/routine.py
@@ -759,19 +759,25 @@ class RoutineBase:
 
         Notes
         -----
-        Differences from ANDES: ANDES has no per-routine ``load_json``;
-        the closest analog is :meth:`andes.routines.tds.TDS.run` with
-        ``from_csv=`` for time-series replay. AMS reads scheduling
-        results back into existing :class:`Var` / :class:`ExpressionCalc`
-        storage and marks the routine as converged so subsequent
-        :meth:`get` and :meth:`ams.system.System.report` calls behave
-        as if the routine just solved.
+        Differences from ANDES: ANDES has no direct analog for
+        post-solve result rehydration. The closest related facility is
+        :meth:`andes.routines.tds.TDS.run` with ``from_csv=``, but
+        that *replays* a time-domain trajectory through the integrator
+        rather than restoring solved values into post-solve storage.
+        AMS reads scheduling results back into existing :class:`Var`
+        and :class:`ExpressionCalc` storage and marks the routine as
+        converged so subsequent :meth:`get` and
+        :meth:`ams.system.System.report` calls behave as if the
+        routine just solved.
 
         .. versionadded:: 1.0.13
 
         .. versionchanged:: 1.2.4
            A successful load now sets ``self.converged = True`` and
-           ``self.exit_code = 0``.
+           ``self.exit_code = 0``. Callers that loaded "candidate"
+           results into a scratch system and relied on
+           ``converged == False`` should reset the flag explicitly
+           after the load.
         """
         try:
             with open(path, 'r') as f:
@@ -823,6 +829,12 @@ class RoutineBase:
         routines, the ``Time`` column carries the literal sentinel
         ``"T1"`` and a single row is expected.
 
+        The export/load contract is a single-space join between key
+        and device idx (``f'{key} {dev}'``) with no quoting. Device
+        idxes are assumed not to collide with this convention; if a
+        future case file uses idxes that contain spaces, the
+        export-side header generation needs revisiting too.
+
         Parameters
         ----------
         path : str
@@ -832,6 +844,8 @@ class RoutineBase:
         -------
         bool
             ``True`` if the loading is successful, ``False`` otherwise.
+            On failure the routine's ``converged`` / ``exit_code``
+            flags are not modified.
 
         Notes
         -----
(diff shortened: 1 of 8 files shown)


ams/utils/paths.py
```diff
@@ -357,8 +369,9 @@ def get_export_path(system, fname, path=None, fmt='csv'):
             final_file_name = f"{provided_base_filename}.{target_extension}"
             full_export_path = os.path.join(dir_path, final_file_name)
     else:
-        # No path provided, use current working directory
-        dir_path = os.getcwd()
+        # No path provided. Prefer system.files.output_path when set, else CWD.
+        output_path = getattr(system.files, 'output_path', None)
+        dir_path = os.path.abspath(output_path) if output_path else os.getcwd()
```
leocuisine, line 374: **api-break / docs (Finding 1).** This default-fallback change also flows into `MatProcessor.export_csv` / `export_npz` (which call `get_export_path(self.owner.system, ...)` at `ams/core/matprocessor.py:160` and `:194`). The PR description's `Out of scope: MatProcessor export_csv default change (not coupled to system.files)` line is contradicted by this code path — a user with `output_path='/some/dir'` who today calls `mats.PTDF.export_csv()` will now get the file under `/some/dir` instead of CWD. Either widen the release-notes blurb to mention MatProcessor, or give `get_export_path` an opt-out and have `MatProcessor.export_csv` pass it.

ams/routines/routine.py
```diff
@@ -793,6 +807,113 @@ def load_json(self, path):
...
+                        item.v = sub.iloc[0].to_numpy(dtype=float)
+                    else:
+                        # CSV is one row per slot, columns per device;
+                        # var.v shape is (n_dev, n_slot) → transpose.
+                        item.v = sub.to_numpy(dtype=float).T
+                except Exception as e:
+                    logger.warning(
+                        f"Failed to assign values to '{key}': {e}")
+
+        self.converged = True
```
leocuisine: **correctness-risk (Finding 6).** A multi-period routine fed a *single-period* CSV (one row, `'T1'`) is detected as `single_period=True` and proceeds — `item.v = sub.iloc[0].to_numpy()` then assigns a 1-D `(n_dev,)` array to a `(n_dev, n_slot)` cvxpy Variable, which cvxpy rejects per-key inside the `try/except` at line 911-913 (warn-and-continue). The routine ends up partially populated **but `converged=True` is set anyway** here. Either detect the period-mismatch up-front (cross-check `single_period` against `hasattr(self, 'timeslot')`) and `return False`, or only flip `converged` when at least one assignment succeeded.

ams/routines/routine.py
```diff
@@ -793,6 +807,113 @@ def load_json(self, path):
                         exprc.v = np.array(values)
                     except Exception as e:
                         logger.warning(f"Failed to assign values to exprc '{key}': {e}")
+
+        self.converged = True
```
leocuisine, line 817: **api-break (minor) / docs (Finding 8).** The release notes flag the new `converged=True`-on-success behavior, good. One thing not flagged: a caller that does `routine.load_json(...); assert not routine.converged` (e.g. as a sanity check after loading candidate results into an unsolved scratch system) silently breaks. Worth adding a one-line workaround to the release-notes bullet (`routine.converged = False` after load) so users hitting it have a documented escape.

leocuisine (commented):
Inline anchors for the `bug`/`api-break` items in the top-level review.

leocuisine (commented):
Independent reviewer (fresh-eyes pass). Diff and source read; no project notes consulted. Round-tripped DCOPF and ED locally to validate the load_csv path; the cvxpy `Var.v` setter writes through to `optz.value`, so the loaded `obj.v` and downstream `Expression.v` evaluate sensibly — the round-trip is more sound than it looks at first read.

Findings below. Anything tagged `bug` / `api-break` is also left as an inline comment.

---

### 1. MatProcessor exports silently inherit the new `output_path` default — **api-break / docs**
`ams/utils/paths.py:371-376`

The PR description's "Out of scope" bullet says *"MatProcessor `export_csv` default change (not coupled to `system.files`)"*, but `MatProcessor.export_csv` (and `export_npz`) call `get_export_path(self.owner.system, ...)` (`ams/core/matprocessor.py:160`, `:194`), so the default-fallback change in `get_export_path` **does** affect them. A user with `output_path='/some/dir'` who previously called `mats.PTDF.export_csv()` and got the file in CWD now gets it in `/some/dir`. The release-notes blurb only mentions routine `export_csv` / `export_json`.

Pick one:
- treat MatProcessor as in scope and add it to the v1.2.4 release-notes bullet (and ideally to the test matrix), or
- give `get_export_path` an opt-out and have `MatProcessor.export_csv` pass it.

Either is fine; the current state is "behavior changed for a caller the description claims wasn't touched."

### 2. `Report.write` return value is documented as "absolute path" but isn't always absolute — **docs / nit**
`ams/report.py:104-109,135,217-218`

When `path` is given, `target` comes from `get_export_path` which calls `os.path.abspath` — absolute, consistent with the docstring. When `path is None`, `target = system.files.txt`, which is built as `os.path.join(output_path, output+'.txt')` (`ams/utils/fileman.py:105`); for the `output_path=''` (i.e. `no_output=False`-without-explicit-dir) path this is a *relative* string. Either tighten the prose ("the path the report was written to") or `os.path.abspath(...)` before returning/logging.

### 3. `load_csv` round-trip for `ExpressionCalc` is asserted nowhere — **test-gap / correctness-risk**
`ams/routines/routine.py:903-913`, `tests/io/test_export_csv.py:90-135`

`pi`, `mu1`, `mu2` on DCOPF/ED are `ExpressionCalc`. The setter at `ams/opt/exprcalc.py:100-109` does `self.optz.value = value`, which only succeeds when `optz` happens to be a leaf cvxpy `Variable` (the dual-variable case today). If a future refactor makes `optz` a derived expression for any exprc (very plausible — `ExpressionCalc` is supposed to support compound `e_str`), the `try/except` at `routine.py:911-913` swallows the failure as a `WARNING` and the round-trip silently drops those columns. The tests today only assert `pg.v` — please add `np.testing.assert_allclose(ss2.DCOPF.pi.v, self.ss.DCOPF.pi.v, atol=1e-5)` (and the same for ED) so any regression here is caught.

### 4. The advertised "load → report" workflow has no test — **test-gap**
`tests/io/test_export_csv.py`, `tests/io/test_report.py`

Both new methods set `converged=True` *specifically so* `system.report()` works on a loaded routine. Neither `test_load_csv_*` nor `test_load_json_*` actually calls `report()` after the load. Smoke-tested locally and it does work (because `Var.v` setter writes through to `optz.value`, which makes `Expression.v` and `obj.v` evaluate from the loaded values), but this is exactly the "if a refactor breaks the writeback path, nothing fails" situation. One assertion of `ss2.DCOPF.report() is True` plus `os.path.exists(...)` on the produced report file would lock it in.

### 5. Failure-mode coverage missing — **test-gap**
`ams/routines/routine.py:852-876`

Three documented `return False` branches have no test:

- `pd.read_csv` raise (e.g. nonexistent path)
- missing `Time` column (`routine.py:858-860`)
- multi-period CSV loaded into a routine without `timeslot` (`routine.py:872-876`)

Also untested: a multi-period CSV with **duplicate** Time entries (e.g. two `'EDT1'` rows). `df.set_index('Time').reindex(slots)` raises `InvalidIndexError: Reindexing only valid with uniquely valued Index objects`. Currently uncaught — it'll bubble out of `load_csv` instead of returning `False`. Either dedupe / detect duplicates explicitly with a clean error, or wrap the multi-period block in the same try/except shape used elsewhere.

### 6. Reverse single↔multi mismatch is not handled symmetrically — **correctness-risk**
`ams/routines/routine.py:868-876`

A *multi-period* routine fed a *single-period* CSV (one row, `'T1'`) is treated as `single_period=True` and proceeds — `item.v = sub.iloc[0].to_numpy()` then assigns a 1-D `(n_dev,)` array to a `(n_dev, n_slot)` cvxpy Variable, which cvxpy will reject. The reverse (single-period routine fed a multi-period CSV) hits the `not hasattr(self, 'timeslot')` branch and returns False cleanly, but the first direction silently warns per-key and leaves the routine partially populated **with `converged=True` set at the end anyway** (`routine.py:915`). Either:
- after the per-key loop, check whether *any* assignment succeeded before flipping `converged`, or
- detect the period-mismatch up front (`single_period` vs `hasattr(self, 'timeslot')`) and fail fast with a clear error.

### 7. `load_csv` columns are matched by exact `f'{key} {dev}'` — **correctness-risk / nit**
`ams/routines/routine.py:890-894`

Device idxes that contain a space (rare but legal — e.g. user-defined `idx='Bus 1'`) will produce a column header like `'pg Bus 1'` on export, and on load the synthesized column name `f'{key} {dev}'` happens to round-trip *only* because both sides use the same join. Worth a comment that the export/load contract is "single-space join, no quoting" — or switch to a safer separator. Not a blocker; flagging because the export/load are now coupled by an unstated convention.

### 8. `load_json` behavior change is technically an api-break — **api-break (minor) / docs**
`ams/routines/routine.py:746-814`

The release notes flag this, good. One thing that's *not* flagged: a caller that today does `routine.load_json(...); assert not routine.converged` (e.g. as a sanity check after loading "candidate" results into an unsolved scratch system) will now break silently. Consider adding to the release notes a one-liner workaround (`routine.converged = False` after load) so users hitting it have a documented escape.

### 9. "Differences from ANDES" notes on `load_json` / `load_csv` — **docs / nit**
`ams/routines/routine.py:760-768`, `:842-849`

Comparing to `andes.routines.tds.TDS.run(from_csv=...)` is a stretch — `from_csv` *replays* a TDS trajectory into the integrator, it doesn't re-hydrate solved values into post-solve storage. The closer ANDES analog is `System.from_csv` / loading a `_out.lst+.npy` pair via `andes.run(..., load_simulation_data=...)`. Consider weakening the analogy ("ANDES has no direct analog; see `TDS.run(from_csv=...)` for a related but distinct replay workflow").

### 10. `gather_results` 6-decimal rounding documented in `load_csv` but not `export_csv` — **docs / nit**
`ams/routines/routine.py:838-840`, `:964-981`

The lossy-at-1e-6 caveat is in the *load* docstring but not the *export* docstring. Users designing a workflow read `export_csv` first; please mirror the note there too (or make it a `gather_results` note that both sides reference).

### 11. `pf` on `report.py:158` — pre-existing but worth noting
`ams/report.py:158`

`rtns_to_collect = [rtn for rtn in system.routines.values() if rtn.converged]` — exactly the loop the new `converged=True` flag in `load_csv` / `load_json` is designed to enable. Worth a brief comment somewhere ("`Report.write` includes any routine with `converged=True`, including those populated via `load_*`") so the coupling between the two changes is documented.

---

## What was checked
- `gh pr view 261`, `gh pr diff 261`
- `ams/report.py`, `ams/system.py`, `ams/utils/paths.py`, `ams/utils/fileman.py`, `ams/routines/routine.py`, `ams/core/matprocessor.py`, `ams/opt/{var,objective,expression,exprcalc}.py`, `ams/models/timeslot.py`, `ams/interface.py`, `ams/io/__init__.py`
- `tests/io/test_export_csv.py`, `tests/io/test_export_json.py`, `tests/io/test_report.py`, `tests/core/test_matprocessor.py`
- `docs/source/release-notes.rst`
- Empirical round-trip: `DCOPF.export_csv → load_csv → report(path=...)` produces the expected report file with correct objective and `pi` values.

## What looks good
- `(n_slot rows × n_dev cols)` CSV layout ↔ `var.v (n_dev, n_slot)` transpose is correct (ED test confirms).
- `'T1'` sentinel detection (`len==1 and time_vals[0]=='T1'`) is robust to numeric-vs-string CSV reads because `'T1'` is non-numeric.
- `path=None` precedence in `system.report` and `Report.write` is consistent.
- `get_export_path` directory-vs-file detection (`splitext` check) handles trailing-slash dirs and explicit-extension files cleanly.
- Tests are placed correctly under `tests/io/` and follow existing module conventions.
- `output_path=''` (FileMan default) is still treated as "not set" in the new branch — backwards-compat for the bare `no_output=True` setup is preserved.

ams/utils/paths.py
leocuisine: Fixed in 4772c78a — the v1.2.4 release-notes bullet now explicitly calls out that `MatProcessor.export_csv` / `export_npz` share the helper and inherit the new `output_path` default. Treating MatProcessor as in scope was the cleaner option (kept `get_export_path` simple).

ams/routines/routine.py
leocuisine: Fixed in 4772c78a. `load_csv` now fails fast in **both** mismatch directions: a single-period CSV into a multi-period routine returns False and leaves `converged` untouched (mirroring the existing multi→single branch). The duplicate-Time pandas `InvalidIndexError` you flagged in Finding 5 is also caught up front. New tests cover both directions and the duplicate-Time case.

ams/routines/routine.py
leocuisine: Done in 4772c78a — release notes now show the explicit opt-out (`ss.DCOPF.converged = False` after the load) so anyone hitting the regression has a one-line workaround documented.

leocuisine:
Thanks for the fresh-eyes review. Fix-up at 4772c78a addresses the substantive findings; itemized below.

**Addressed**

- **#1 MatProcessor.** Treated as in scope — release notes now explicitly call out that `MatProcessor.export_csv`/`export_npz` inherit the new `output_path` default. Cleaner than adding an opt-out kwarg.
- **#2 Report.write return.** `target = os.path.abspath(system.files.txt)` in the `path is None` branch so the documented "absolute path" return value holds for both branches.
- **#3 ExpressionCalc round-trip.** `test_load_csv_DCOPF` now asserts `pi`, `mu1`, `mu2` (and ED asserts `pi`). The empirical observation that `Var.v` writes through to `optz.value` is now locked in by tests.
- **#4 Load → report workflow.** `test_load_csv_DCOPF` now produces a report file from the loaded routine and asserts it exists. The headline use case is covered.
- **#5 Failure-mode coverage + duplicate Time bug.** Four new tests: missing `Time` column, both period-mismatch directions, and duplicate Time entries. The duplicate-Time `InvalidIndexError` is now caught up front in `_load_csv_align_period` instead of bubbling out of `pd.reindex`.
- **#6 Period-mismatch symmetry.** `load_csv` is split into `_load_csv_align_period` + `_load_csv_assign`; the align step fail-fasts in **both** directions (single-into-multi as well as multi-into-single) and `converged`/`exit_code` are only set after the assign step returns. No more silent corruption + `converged=True`.
- **#8 load_json workaround.** Release notes now show `ss.DCOPF.converged = False` as the explicit opt-out for callers that relied on the old behavior.
- **#9 ANDES analogy.** Both `load_json` and `load_csv` Notes now state ANDES has no direct analog and weaken the `TDS.run(from_csv=...)` reference to "related but distinct — replays a trajectory rather than rehydrating post-solve storage."
- **#10 Rounding caveat.** `export_csv` docstring now mirrors the load-side note about 6-decimal export rounding.

**Codacy follow-ups bundled in the same commit**

- `load_csv` complexity dropped from 16 → under threshold by the split above (Codacy's 1 medium complexity finding).
- All new logger calls in `load_csv` switched to lazy `%`-formatting to match the `ams/system.py` convention (likely the 7 minor BestPractice findings).

**Acknowledged, not actioned this PR**

- **#7 Device idxes with spaces.** Documented the contract ("single-space join, no quoting") in the `load_csv` docstring. Switching to a safer separator would be an export-format change worth its own discussion; flagging here so future-me has the trail.
- **#11 `rtns_to_collect` comment.** Pre-existing; left for a separate readability sweep.

Local checks: 55/55 io tests green (4 new), 372/372 full suite, flake8 clean on touched files (no new C901s).

Issue #164, opened 2025-05-23
In model EDTSlot and UCTSlot, make the mapping explicit


jinningwang:
**Is your feature request related to a problem? Please describe.**

In model EDTSlot and UCTSlot, the mapping relationship between loading factors and target devices are unclear.

**Describe the solution you'd like**

Maybe add two new params in these two model, one for model or group name, and another for time series values.

**Describe alternatives you've considered**

**Additional context**

jinningwang:
A possible implementation of timeseries loading factors:

| uid | idx | Area | sd |
|:---:|:---:|:---:|:---:|
| 1 | 1 | 1 | [1, 0.9, 0.8] |
| 2 | 2 | 2 | [1, 0.9, 0.8] |

jinningwang:
A possible implementation of commitment decision model:
| uid | idx | gen | ud |
|:---:|:---:|:---:|:---:|
| 1 | 1 | 1 | [1, 0, 1] |
| 2 | 2 | 2 | [1, 0, 1] |

jinningwang:
The models in above comments actually move the implicit mapping to `TimeSlot` from generators and areas.

Then how do we index the time slot explicitly?

jinningwang:
A possible simpler solution is to explain this in the documentation, like "How do I know the order of generators? You can print out pg.get_all_idxes, and that is the order of `ug` in models `EDTSlot` and `UCTSlot`"

Sections of interests can include:
- Models `EDTSlot` and `UCTSlot`
- Routines `ED` `UC` and their derivatives 
- Examples of using `ED` and `UC`

jinningwang:
However, it is better to use implicit mapping to TimeSlot, since it is more intuitive as time naturally increases.

Then we need:
1. a new model `TimeSlot2` where it has parameters `u` `idx` `name`.
1. new routines `ED` `UC` etc. need to 1) have a new `RParam` named `timeslot`, 2) assign necessary variables `horizon` attribute to this parameter.
1. a new model `UCT2EDT` to store the mapping relationship between UC results to ED results. This will be used in the interface methods between UC and ED https://github.com/CURENT/ams/issues/169

PR #262, opened 2026-05-03
feat: TimeSlot paired-key data model + RParam.horizon (v1.3.0)

leocuisine:
## Summary

Replaces the pre-v1.3.0 single-table CSV-list TimeSlot encoding with explicit paired-key tables, plus a small extension to `RParam` so input 2D parameters mirror the `Var.horizon` shape AMS already uses on the output side.

**This is a breaking case-file change targeting v1.3.0.** The unreleased v1.2.4 label was retired into v1.3.0 (rolling in the report-path + load_csv work from #261).

### What changed

1. **`RParam.horizon` + `hindexer`** (`ams/core/param.py`) — input 2D pivot, mirroring `Var.horizon`. ` RParam(model='EDSlotGen', indexer='gen', imodel='StaticGen', horizon=self.timeslot, hindexer='slot')` returns a `(ngen, nslot)` matrix on `.v`. Cells with no matching row default to `NumParam.default`; duplicate `(primary, secondary)` rows raise.

2. **TimeSlot model split** (`ams/models/timeslot.py`) — definitions vs. data:
   - `EDTSlot` → `EDSlot` (slot definitions only)
   - `UCTSlot` → `UCSlot` (slot definitions only)
   - new `EDSlotLoad` per-`(area, slot)` `sd`
   - new `EDSlotGen` per-`(gen, slot)` `ug`
   - new `UCSlotLoad` per-`(area, slot)` `sd`
   - unused `GCommit` stub deleted

3. **Routine rewiring** (`ams/routines/{ed,uc}.py`) — drops the `ED.ugt = NumOp(np.transpose, …)` workaround; the new `EDSlotGen` shape produces `(ngen, nslot)` directly. The `MPBase.sdT` back-compat shim was eliminated by flipping `LoadScale.v` to consume sd as `(narea, nslot)` and routing the `SRBase`/`NSRBase` reserve chain through a small `multiply_left_t` helper.

4. **`tools/migrate_timeslot.py`** — JSON + XLSX, idempotent, in-place with `.bak`. Detects legacy by cell content (comma-string), no version stamp needed. Refuses partly-migrated input rather than risk silent data loss. The 9 case files in `ams/cases/` are migrated.

5. **Routine model-ref validator** (`Routine._validate_model_refs`) — runs at the top of `_data_check`; walks every RParam/Var, resolves `model`/`imodel` against `system.models`/`system.groups`/`'mats'`, raises with a `difflib.get_close_matches` suggestion on unresolved refs. Catches typos like `model='EDSlotLooad'` that would previously silently land as `rparam.owner is None` and skip.

### Why (breaking change)

The legacy `EDTSlot.ug` / `EDTSlot.sd` / `UCTSlot.sd` cells were CSV-list strings whose **list position implicitly aligned** with `StaticGen.get_all_idxes()` / `Area.get_all_idxes()`. This implicit contract was the original [#164](https://github.com/CURENT/ams/issues/164) report and matched no other AMS data convention; explicit paired keying eliminates a class of silent ordering bugs.

### Migration

```sh
python tools/migrate_timeslot.py path/to/case.json --in-place
python tools/migrate_timeslot.py path/to/case.xlsx --in-place
```

A `.bak` copy is saved alongside. The migrator is idempotent — safe to re-run.

### Validation

- 390/390 tests pass (was 372 on develop; +18 new tests across `RParam.horizon` pivot, migrator, and validator)
- flake8 clean
- sphinx docs build green
- Solver-level verification on pjm5bus_demo: ED objective unchanged pre/post de-shim (`433.3668`); UC solves to `430.8604` with SCIP

### Pre-PR review

A local fresh-eyes code review was done before pushing — 21 findings triaged, 11 fixed (including 1 blocker on partial-state migrator behavior), 10 skipped with rationale. Fixes batched in commit `8ec5c2fc`.

## Test plan

- [ ] CI green across all 9 cells (ubuntu/macos/windows × 3.11/3.12/3.13)
- [ ] Codacy green (or known acceptable)
- [ ] ReadTheDocs build green
- [ ] Independent reviewer agent comments addressed


tools/migrate_timeslot.py
```diff
@@ -0,0 +1,453 @@
...
+    for slot in slot_rows:
+        if 'ug' not in slot:
+            continue
+        if not _is_legacy_cell(slot['ug']):
+            continue
+        parsed = _parse_csv_list(slot['ug'])
+        if len(parsed) != len(gens):
+            raise ValueError(
+                f"EDSlot: slot {slot.get('idx')!r} has ug of "
+                f"length {len(parsed)}, expected {len(gens)} "
```
leocuisine, line 183: Finding #1 (bug). The PR description says "Duplicate slots in the source raise" but neither this helper nor migrate_data() ever checks EDTSlot[*].idx / UCTSlot[*].idx for duplicates. Two EDT1 rows with different sd cells will both expand into (area, slot=EDT1) rows here, and the failure surfaces only later at RParam._materialize_2d first-.v access — with no source-row context. Either drop the docstring claim or add an explicit duplicate-slot check (with offending idx + source row positions) at the top of migrate_data, alongside the partly-migrated guard.

docs/source/release-notes.rst
```diff
@@ -6,12 +6,79 @@ Release notes
...
+matching the e_str expectations. Likewise, the ``MPBase.sdT``
+shim was eliminated by flipping
+:meth:`ams.core.service.LoadScale.v` to consume sd as
+``(narea, nslot)`` and routing the SRBase / NSRBase reserve chain
+through a small :func:`ams.utils.func.multiply_left_t` helper.
+
+Note that this is a behavior change for any user code calling
+:class:`ams.core.service.LoadScale` directly with a custom ``sd``
+RParam: the expected input shape is now ``(narea, nslot)``, not
+``(nslot, narea)``. Likewise, ``EDSlot`` / ``UCSlot`` no longer
```
leocuisine, line 93: Finding #5 (api-break). The "EDSlot / UCSlot no longer carry sd / ug columns" line is buried inside the LoadScale-shape paragraph, but it is its own user-facing AttributeError for anyone reading ss.EDSlot.sd.v / ss.EDTSlot.sd.v directly in user code. Lift it into its own bullet under the "Breaking" section near the model rename, alongside the migrator pointer.

ams/routines/routine.py
```diff
@@ -540,10 +541,62 @@ def _get_off_constrs(self):
             logger.warning(msg)
         return disabled
 
+    def _validate_model_refs(self):
```
leocuisine, line 544: Finding #6 (api-break). _validate_model_refs is itself a backward-incompat: any user routine subclass that referenced a typo-d model and skated by under the old silent-skip behavior will now hard-error at init(). The release notes do not call this out. Likely fine in practice (the silent skip was always a bug), but worth a one-liner under v1.3.0 release notes so users debugging an init() regression can find the rationale.

leocuisine (commented):
Independent review of the TimeSlot paired-key + RParam.horizon work
(commits `34315c14..8ec5c2fc`). I read `ams/`, `tests/`, `tools/`,
`docs/source/release-notes.rst`, and the migrated case files; I did
not read anything under `.claude/` (or the projects/memory notes) so
my read of the design is from the diff alone. I did not execute the
test suite — comments below are static-analysis findings, not
solver-level reproductions.

Headline assessment up front: this is a well-scoped breaking change
with thoughtful surfacing (loud migrator, validator, in-tree case
files all migrated). The numerical-shape bookkeeping checks out:
`transpose(multiply(a,b)) == multiply(transpose(a), transpose(b))`
combined with `sd` going from `(nslot, narea)` to `(narea, nslot)`
makes the `multiply_left_t` substitution and the new
`LoadScale[u_yloc, :]` indexing algebraically equivalent to the
pre-refactor chain. No blocker findings.

---

## bug

**1. `tools/migrate_timeslot.py:148-152` — duplicate slot idxes silently produce duplicate emitted rows.** The PR description states "Duplicate slots in the source raise" but the migrator never checks for repeats in `EDTSlot[*].idx` / `UCTSlot[*].idx`. Two `EDT1` rows with different `sd` cells will both expand into `(area, slot=EDT1)` rows, and the failure surfaces only later at `RParam._materialize_2d` first-`.v` access — with no file or line context to help the case author diagnose. Either drop the docstring claim or add an explicit duplicate-slot check at the top of `migrate_data` and raise with the offending idx + source row positions. Cheap to add and consistent with the partly-migrated guard already in place.

## correctness-risk

**2. `tools/migrate_timeslot.py:165-189` (rename-only path) — definition rows are not slimmed.** When `_rename_legacy_keys` returns `renamed=True` but `has_legacy_cells=False` (single-area case where `sd: "0.5"` doesn't trip the comma test, or someone hand-removed the CSV cells but left empty `sd` / `ug` columns), the function returns at `if not has_legacy_cells: return` without ever calling `_slim_definition_row`. The renamed `EDSlot` table then still carries `sd` / `ug` columns the new schema doesn't define. Whether AMS load tolerates the extra columns depends on the loader's strictness; either way it's an unobvious non-invariant. Suggest unconditionally slimming definition rows when iterating.

**3. `tools/migrate_timeslot.py:46-56` — `_is_legacy_cell` mis-classifies single-area cells.** A single-area case with `sd: "0.5"` has no comma → `_is_legacy_cell` returns False → the row is kept on the (now-misnamed) definition table. None of the nine in-tree cases exercise this (all have ≥3 areas), but external case authors writing minimal one-area scenarios will silently mis-migrate. Either accept any string-or-numeric `sd`/`ug` cell on a legacy table as "legacy" (and rely on the slim/emit-rows path to handle it), or document the lower bound on areas explicitly in the migrator usage.

**4. `tools/migrate_timeslot.py:107-119` (`_emit_gen_rows`) — silently drops `ug` data when `Slack` and `PV` blocks are both absent.** The helper hard-codes `('Slack', 'PV')` (in `_device_orderings`) for membership of `StaticGen`. If a custom case ships gens via a different group ordering, the dropped data goes only to a warning. Either upgrade to error (the docstring rationale "drop unloadable rows" is much weaker for `ug` than for `sd` — empty gen list + non-empty `EDTSlot[*].ug` is almost certainly a bug, not a quirk), or teach `_device_orderings` about the AMS group-membership rule rather than hard-coding model names inline.

## api-break

**5. `docs/source/release-notes.rst:62-80` — release notes call out the `LoadScale` shape change but bury the `EDSlot.sd` / `EDSlot.ug` removal as an attribute break.** Anyone reading `ss.EDSlot.sd.v` or `ss.EDTSlot.sd.v` directly in user code will get an `AttributeError` after upgrading. The "EDSlot / UCSlot no longer carry sd / ug columns" line is buried inside the LoadScale paragraph. Lift it into its own bullet under the "Breaking" section near the model rename, alongside the migrator pointer.

**6. `ams/routines/routine.py:544-595` — the new `_validate_model_refs` is itself a backward-incompat:** any user-defined routine subclass that referenced a typo'd model and skated by previously will now hard-error at `init()`. Likely fine in practice (the silent skip was always a bug), but it's worth a one-liner under the v1.3.0 release notes so users debugging an `init()` failure can find the rationale.

## test-gap

**7. `tests/test_migrate_timeslot.py` — no XLSX round-trip test.** All migrator tests exercise `migrate_data` on Python dicts. The XLSX adapter (`_xlsx_load`, `_xlsx_dump`, `_header_sort_key`, the sheet-order rename mirroring at `migrate_xlsx:381-395`) has zero direct coverage and is the part most likely to regress when openpyxl semantics shift. A single end-to-end test that writes a tiny xlsx with `EDTSlot` / `UCTSlot` sheets, runs the migrator, and re-loads with openpyxl to assert sheet names + per-row content would close this.

**8. `tests/test_migrate_timeslot.py` — no idempotence test on a fully-migrated case from disk.** `test_idempotent_on_migrated_input` uses `deepcopy(out1)` of an in-memory result. The on-disk JSON written by `migrate_json` (with `indent=2`, trailing newline) round-tripping through a second invocation has no test, even though the migrator advertises `--in-place` as safe to re-run.

**9. `tests/core/test_rparam_horizon.py` — no test for `expand_dims` interaction with `horizon`.** `RParam._v` evaluates `expand_dims` after the 2D pivot returns. If a future user sets both, the result silently goes 3D. Not a current bug, but a one-line assertion would document the intended behavior.

**10. `tests/routines/test_routine_base.py:42-47` — `test_validate_model_refs_mats_whitelist` only verifies `'mats'` doesn't raise; it does not verify that an invalid value (e.g. `'matz'`) is suggested as `'mats'`.** The difflib suggestion path is the value-add of the validator and is otherwise untested.

## docs

**11. `docs/source/release-notes.rst:62-67` — the "Internal" header for the de-shim work understates the user-facing impact.** The LoadScale shape contract is explicitly an external API break for any user code that constructed a `LoadScale` ROperationService themselves. Retitle to "Behavior change" and fold the `(narea, nslot)` requirement up into a dedicated bullet.

**12. `ams/core/param.py:51-67` — `horizon=` docstring says cells with no matching row "fall back to the source `NumParam.default`" but `_materialize_2d` does `getattr(src_param, 'default', 0)`,** falling through to `0` (not the source default) if the source has no `default`. For the in-tree `EDSlotLoad.sd` (default=1.0) and `EDSlotGen.ug` (default=1) this is irrelevant, but the docstring is wrong about the fallback chain. Either fix the prose to call out the `0` fallback or drop the `, 0` second arg from `getattr` so the docstring is accurate.

**13. `ams/utils/func.py:90-106` — `multiply_left_t` is documented in terms of "the v1.3.0 sd shape"** — useful context, but the function itself is generic. Lead with the math (`Returns np.multiply(a.T, b)`) and follow with the routine-specific use case.

## style

**14. `ams/core/param.py:228-249` — `_materialize_2d` builds `seen` only to guard against duplicates, then re-iterates `seen.items()` to populate `out`.** The duplicate check could happen inline in the second loop, dropping the dict. Minor; current code is more debuggable.

**15. `ams/routines/routine.py:585-593` — the unresolved-refs raise uses bare `ValueError`** with no class hint that this is a model-resolution failure. Consider a custom exception subclass or at least prefix the message with `RoutineModelRefError:` so wrappers / `except ValueError` blocks elsewhere can disambiguate.

**16. `tools/migrate_timeslot.py:343-359` — `_xlsx_dump` recomputes `_header_sort_key(rows)` once per sheet and wraps `sorted(set, key=…)`** — `O(headers²)` if a sheet ever grows wide. Not worth fixing now.

## nit

**17. `ams/routines/ed.py:140-153` — the comment block reads naturally but the "doesn't see a double-registration" rationale would be more discoverable if it pointed at `addRParam` (the method that the late-bind sidesteps).**

**18. `tools/migrate_timeslot.py:99-101, 117-119` — `head = legacy[:3] + (['...'] if len(legacy) > 3 else [])` is a clever one-liner for a truncated preview** but the `+` against a list literal will surprise readers expecting a string-formatting helper.

**19. `ams/models/timeslot.py:24-93` — three near-identical `Model.__init__` blocks with `self.group = 'Horizon'`.** A small mixin or helper would dedupe; current shape is fine and easy to read.

---

## Overall assessment

This branch is in good shape to merge after CI green. The numerical-shape rewiring is internally consistent (the `transpose-of-multiply` identity is preserved end-to-end), the migrator is conservative (refuses partly-migrated state), and the new validator catches a real class of silent bugs. The findings above are dominated by migrator edge cases and docs polish — none should block. The single recurring pattern worth noting is in `tools/migrate_timeslot.py`, where a few code paths advertise behavior (duplicate-slot detection, `NumParam.default` fallback in the param docstring) that the implementation doesn't quite deliver; tightening these is cheap and worth doing before v1.3.0 ships externally.

tools/migrate_timeslot.py
leocuisine: Fixed in 9cdb2109 — `migrate_data` now does an explicit duplicate-slot pass at the top, raising `ValueError` with the offending idx + source row positions before the per-axis emission can hit the inner `RParam._materialize_2d` failure path. New regression test at tests/test_migrate_timeslot.py::TestDuplicateSlotGuard::test_duplicate_slot_idx_raises.

docs/source/release-notes.rst
leocuisine: Fixed in 9cdb2109 — `EDSlot/UCSlot no longer carry sd/ug attributes` is now its own bullet under the breaking section, ahead of the LoadScale-shape paragraph. Calls out the `AttributeError` upgrade path explicitly and points readers at `rtn.sd.v` / `rtn.ug.v` for the 2D matrix view.

ams/routines/routine.py
leocuisine: Fixed in 9cdb2109 — added a release-notes bullet calling out that `_data_check` now hard-fails on unresolved `model`/`imodel` references via `RoutineModelRefError`. Also prefixed the `ValueError` message with `RoutineModelRefError:` so wrappers that `except ValueError` elsewhere can disambiguate without subclassing.

leocuisine:
Triaged the 19 findings — 13 fixed in `9cdb2109`, 6 skipped per the reviewer's own "not-worth-fixing-now" / nit annotations.

**Fixed (13):**
- **bug #1** — duplicate-slot guard now actually implemented in `migrate_data` (was docstring-only); raises with offending idx + row positions, regression test added.
- **correctness #2** — rename-only path now slims definition rows so vestigial empty `sd`/`ug` columns don't survive the migration.
- **correctness #3** — `_is_legacy_cell` gains a `force=` flag set true when the source key was renamed; catches single-area cases whose `sd` is bare-numeric `"0.5"` with no comma.
- **correctness #4** — empty `StaticGen` + non-empty `ug` now raises rather than warns (left the parallel `Area`+`sd` path as warn-and-drop, since npcc/wecc legitimately exhibit that pre-v1.3.0 quirk).
- **api-break #5/#6** — release notes now have dedicated bullets for the `EDSlot.sd`/`EDSlot.ug` attribute removal and the new `RoutineModelRefError` hard-fail in `_data_check`. The "Internal" header on the de-shim section is retitled "Behavior change" since the LoadScale shape contract is genuinely external (#11).
- **test-gap #7/#8/#10** — added end-to-end XLSX round-trip test (sheet rename + per-axis emission verified by openpyxl re-load), on-disk JSON idempotence test, and explicit difflib-suggestion assertion in the validator test.
- **docs #12** — dropped the `, 0` fallback in `getattr(src_param, 'default', 0)` so the actual behavior matches the docstring claim of `NumParam.default` fallback (better than weakening the docs).
- **docs #13** — `multiply_left_t` docstring now leads with the math signature, version-specific use case follows.
- **style #15** — error message prefixed with `RoutineModelRefError:` for grep / disambiguation.

**Skipped with rationale:**
- #9 (`expand_dims × horizon` test): speculative, no current bug.
- #14 (`_materialize_2d` `seen` dict refactor): your own note "current is more debuggable" — agreed.
- #16 (`_xlsx_dump` O(headers²)): your own note "not worth fixing now".
- #17 / #18 / #19 (nits): your own note "current shape is fine and easy to read".

Test count: 372 → 395 (+23). Lint clean, sphinx docs build green.

PR #264, opened 2026-05-03
refactor(service): deprecate NumExpandDim/VarReduction; share scalar-wrap helper

leocuisine:
## Summary

- **Deprecate** `NumExpandDim` and `VarReduction`. Both classes have **zero call sites** under `ams/routines/` and `ams/core/`. Their `__init__` now emits `DeprecationWarning` (`stacklevel=2`); docstrings carry `.. deprecated:: 1.3.0`. Removal target is **v1.4.0**.
- **Extract** `_ensure_array(out)` as a module-level helper so `NumOp.v0` and `NumOpDual.v0` stop carrying near-duplicate scalar→ndarray wrap blocks.

Phase 2 + Phase 3 of `projects/service_module_cleanup/`. Phase 1 (docs + structure) shipped in #263.

## Replacements

| Deprecated form | Replacement |
|---|---|
| `NumExpandDim(u, axis=N)` | `NumOp(u, fun=np.expand_dims, expand_dims=N)` |
| `VarReduction(u, fun=np.ones)` | inline: `np.ones((1, u.n))` |

## Behavior preservation

`_ensure_array` is byte-equivalent to the inline blocks that previously lived in `NumOp.v0` / `NumOpDual.v0` (scalar → 1-element ndarray; lists/tuples → `np.asarray`; ndarray and scipy.sparse passthrough). `NumOpDual.v0` still applies its post-wrap sparse conversion as before — that quirk is left alone (out of scope; harmless because `csr_matrix(csr_matrix(x))` is idempotent).

## Test plan

- [x] `pytest tests/core/test_service.py -v` — 10 pass (was 7), including 3 new deprecation tests:
  - `test_NumExpandDim_warns_on_init` — `pytest.warns(DeprecationWarning, match="NumExpandDim")`
  - `test_VarReduction_warns_on_init` — `pytest.warns(DeprecationWarning, match="VarReduction")`
  - `test_module_import_does_not_warn` — guard that bare `import ams.core.service` does not fire either warning
- [x] `pytest tests/` — 398 passed
- [x] `cd docs && make html SPHINXOPTS="-W --keep-going"` — clean

### DIFF
--- a/ams/core/service.py
+++ b/ams/core/service.py
@@ -25,8 +25,10 @@ UC min on/off duration        :class:`MinDur`          (UC routines only)
 Notes
 -----
 - The ``expand_dims=axis`` kwarg on :class:`NumOp` covers the
-  :class:`NumExpandDim` use case; ``NumExpandDim`` is retained as a
-  thin alias for backward compatibility — prefer ``NumOp`` directly.
+  :class:`NumExpandDim` use case. ``NumExpandDim`` is **deprecated
+  in v1.3.0** and slated for removal in v1.4.0 — use ``NumOp``
+  directly. :class:`VarReduction` is likewise deprecated (no
+  production users); inline the reduction matrix instead.
 - :class:`MinDur` inherits from :class:`NumOpDual` for parameter
   plumbing only; the inherited ``fun``/``rfun`` machinery is unused.
 - ``.v`` caching: subclasses recompute fresh on every access — there
@@ -47,6 +49,7 @@ File layout
 """
 
 import logging
+import warnings
 from typing import Callable, Type
 
 import numpy as np
@@ -58,6 +61,22 @@ from ams.opt import Param
 
 
 logger = logging.getLogger(__name__)
+
+
+def _ensure_array(out):
+    """
+    Coerce a service output to an ndarray-like for downstream parsing.
+
+    Genuine Python scalars become 1-element ndarrays; ``ndarray`` and
+    ``scipy.sparse`` matrices pass through untouched; other array-likes
+    (list, tuple, ...) go through :func:`numpy.asarray`. Used by
+    :class:`NumOp` / :class:`NumOpDual` ``v0`` when ``array_out=True``.
+    """
+    if isinstance(out, np.ndarray) or spr.issparse(out):
+        return out
+    if np.isscalar(out):
+        return np.array([out])
+    return np.asarray(out)
 
 
 # ---------------------------------------------------------------------------
@@ -306,14 +325,7 @@ class NumOp(ROperationService):
     def v0(self):
         out = self.fun(self.u.v, **self.args)
         if self.array_out:
-            # wrap genuine Python scalars in a 1-element ndarray; pass
-            # ndarrays and scipy.sparse matrices through untouched;
-            # convert other array-likes (lists/tuples) via np.asarray
-            if not isinstance(out, np.ndarray) and not spr.issparse(out):
-                if np.isscalar(out):
-                    out = np.array([out])
-                else:
-                    out = np.asarray(out)
+            out = _ensure_array(out)
         return out
 
     @property
@@ -475,14 +487,7 @@ class NumOpDual(NumOp):
     def v0(self):
         out = self.fun(self.u.v, self.u2.v, **self.args)
         if self.array_out:
-            # wrap genuine Python scalars in a 1-element ndarray; pass
-            # ndarrays and scipy.sparse matrices through untouched;
-            # convert other array-likes (lists/tuples) via np.asarray
-            if not isinstance(out, np.ndarray) and not spr.issparse(out):
-                if np.isscalar(out):
-                    out = np.array([out])
-                else:
-                    out = np.asarray(out)
+            out = _ensure_array(out)
         if self.sparse:
             return spr.csr_matrix(out)
         return out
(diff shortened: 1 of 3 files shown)


leocuisine (commented):
**Independent reviewer agent — fresh-eyes pass on the diff and source.**

What I checked: behavior parity of `_ensure_array` against the pre-PR inline blocks; `stacklevel=2` correctness; the zero-call-sites claim across `ams/`, `tests/`, `examples/`, `docs/source/`; the three new tests; release-notes RST; preservation of `NumExpandDim.v` / `VarReduction.v0` semantics; consistency with the existing `@deprecated` pattern in `ams/utils/misc.py`.

**Findings**

1. `test-gap` — `tests/core/test_service.py:105-129`. The three new tests pin only the warning contract; they never instantiate-then-evaluate. A user who silences the `DeprecationWarning` still relies on `NumExpandDim(...).v` and `VarReduction(...).v0` returning the same arrays they did pre-PR. Suggest one extra assertion per class, e.g.:

   ```python
   with warnings.catch_warnings():
       warnings.simplefilter("ignore", DeprecationWarning)
       svc = NumExpandDim(u=rp, axis=0)
   np.testing.assert_array_equal(svc.v, np.expand_dims(rp.v, axis=0))
   ```

   Cheap insurance that the deprecation didn't accidentally regress output, and pins the contract for the v1.4.0 removal review.

2. `nit` — `ams/core/service.py:20`. The "Service selection guide" table still lists `VarReduction` as the recommended choice for "Shape-only reduction matrix". With v1.3.0 deprecating it, the table row should either be removed or annotated (e.g. drop the row entirely, since the replacement is `np.ones((1, u.n))` inline — there is no recommended class). Same for the "8. Backward-compat alias (`NumExpandDim`)" line at `:48` — the file-layout comment now misleads, since the class is being removed in v1.4.0, not retained as an alias.

3. `nit` — `ams/utils/misc.py:115` already defines a `deprecated()` decorator (used by `deprec_get_idx`). The two new `__init__` warnings could plausibly use the same helper for consistency (e.g. wrap `__init__` with `@deprecated("NumExpandDim is deprecated…")`). Not blocking — the manual `warnings.warn` is also fine and arguably clearer for `__init__` — but worth a one-line acknowledgment in the PR if the author considered and rejected it.

4. `nit` — release-notes bullets at `docs/source/release-notes.rst:172,175` use the `=…` ellipsis character (`…`, U+2026). Existing release-notes copy uses ASCII `...`; mixing is harmless but inconsistent with the rest of the file. (Verified that `make html SPHINXOPTS=-W` is reported clean by the author, so this is purely cosmetic.)

**Verified clean**

- `_ensure_array` body is byte-equivalent to the two pre-PR inline blocks (same isinstance/issparse short-circuit ordering, same `np.isscalar`/`np.asarray` branches). 0-d ndarrays and `np.float64` scalars route to the same branch as before.
- `stacklevel=2` is correct: warning surfaces at the user's `NumExpandDim(...)` / `VarReduction(...)` call site, not the `__init__` line.
- Zero-uses claim holds: outside the deprecation site, the new tests, the release-notes bullet, the in-file selection-guide row, and the auto-generated `docs/source/_generated/` autosummary stubs (not authored), neither name appears anywhere under `ams/`, `tests/`, `examples/`, or `docs/source/`.
- `NumExpandDim.v` and `VarReduction.v0` overrides are untouched; the deprecation is purely additive at `__init__`.
- The `test_module_import_does_not_warn` `importlib.reload` pattern is the right way to verify import-time silence; raising on `DeprecationWarning` ensures any module-scope instantiation regression would fail loudly.

No `bug` / `api-break` / `correctness-risk` findings. Approving in spirit pending the test-gap fix.

leocuisine:
Thanks for the careful read. Walking through each finding:

**1. test-gap (extra `assert_array_equal` per deprecated class).** Done in 510f30ba. The two deprecation tests now pin `.v` / `.v0` output equality with the documented replacement (`np.expand_dims(rp.v, axis=0)` for `NumExpandDim`, `np.ones((1, rp.n))` for `VarReduction`). This buys the v1.4.0 removal review a tripwire: if migration parity ever breaks before removal, those tests catch it.

**2. nit (selection guide still references the deprecated classes).** Fixed in 510f30ba. The "Shape-only reduction matrix" row in the guide table now points at the inline `np.ones((1, u.n))` replacement instead of `VarReduction`, and the section-8 layout banner is relabeled "Deprecated — slated for removal in v1.4.0" rather than "Backward-compat alias".

**3. nit (use the in-repo `deprecated()` decorator from `ams/utils/misc.py`).** Considered — declining for now. That decorator is shaped for plain functions/methods (it wraps with `functools.wraps` and forwards `*args, **kwargs`), and applying it to `__init__` of a class that uses keyword-only-style positional args would still produce a working warning but the resulting stack trace points at the wrapper rather than the class line itself. The explicit `warnings.warn(..., stacklevel=2)` shape is the more conventional pattern for class deprecations and gives a cleaner traceback. Happy to revisit if a stronger convention emerges.

**4. nit (U+2026 ellipses).** Fixed in 510f30ba — the four `…` chars in the new release-notes section are now ASCII `...`, matching the dominant style in the file (the one stray U+2026 at line 82 is pre-existing in the TimeSlot section and out of scope here).

All four addressed; no behavior changes from the fix-up. Tests still pass (10/10 in `tests/core/test_service.py`, 398 overall) and `make html SPHINXOPTS="-W"` is clean.

PR #268, opened 2026-06-01
feat(uc): enforce minimum on/off duration + fix status-filtered Cg

leocuisine:
## Summary

Makes the UC family's minimum on/off duration constraints actually take effect, and fixes a core `MatProcessor` bug surfaced in the process.

Before this branch, UC commitment was **frozen in time**: the startup/shutdown coupling used equality constraints that, under binary `vgd`/`wgd`, forced `ugd[g, t] == ug0[g]` for every period. Minimum on/off duration was never enforced, and `StaticGen.ton0`/`toff0` were dead parameters.

## Changes

**UC minimum on/off duration (Phases 0–3):**
- `fix(uc)`: replace the `actv`/`actw` equality coupling with the standard three-binary (Rajan & Takriti) state equation, so commitment can vary across the horizon.
- `feat(uc)`: initial-state minimum on/off duration enforced from `ton0`/`toff0` (elapsed on/off time at `t=0`, hours) via a new `MinDurInit` service.
- `feat(uc)`: interior minimum on/off duration via a Rajan–Takriti window-sum (`MinDurWindow` service, block-diagonal band matrix; `cp.vec(..., order='C')`).
- Phase 0 characterization tests pin the prior stuck-commitment behavior before the rewrite.

**Core fix — `build_cg` ignored generator online status:**
- `MatProcessor.build_cg` filtered `Cg` columns by `StaticGen.u`, zeroing the column of any offline generator — contrary to its documented contract ("Generator online status is NOT considered in its connectivity matrix").
- Because UC optimizes commitment via `ugd` (and `StaticGen.u` only seeds the initial state), a generator offline at matrix-build time — e.g. one turned off by UC's initial-guess heuristic — stayed permanently invisible to the nodal (angle) network even after the optimizer re-committed it.
- This was the root cause of the long-standing UCES vs `UC2ES` objective mismatch: UCES (`Cg@pg`) couldn't use the re-committed unit (obj 425, wrong); `UC2ES` (`cp.sum(pg)`) could (obj 368, right). `Cg` is now structural; offline units stay pinned to zero output through existing capacity bounds. With the fix, UCES == UC2ES (368.2068) on obj/ugd/pg/plf; only UCES changed (425→368), all other UC routines unchanged.

**Tests (`test_scenarios_milp_multiperiod`):**
- Removed both `UC2ES` xfails and the now-unneeded `align_full=False` (UC2ES matches UCES fully).
- Rewrote `test_trip_gen`: its old "warm-start-off gen stays off" assertion was bug-dependent (UC optimizes commitment; a guessed-off unit may legitimately be re-committed). New assertion is the portable invariant: uncommitted gens (`ugd=0`) produce no power.

## Behavior change

UC dispatch and objective values change relative to v1.3.0 for cases where commitment would otherwise vary — the prior results were not enforcing min on/off duration. Documented in the v1.3.1 release notes.

## Validation

Full suite: **410 passed, 0 failed, 0 xfailed** (was 408 passed + 2 xfailed).

## Known follow-ups (out of scope, tracked separately)

- **Vacuous PTDF line limits:** UC2/UC2ES's `plflb`/`plfub`/`alflb`/`alfub` still reference `Bf@aBus`, but the PTDF routines drop the nodal balance that pins `aBus`, leaving the line/angle limits inert. Harmless on uncongested `pjm5bus_demo` but would let a congested case violate limits. Needs its own PR (re-express limits on the PTDF `plf` expression).
- `build_cl`/`build_csh`/`build_cft` also filter on element status (not bugs there, but the same docstring/code mismatch should be reconciled).

### DIFF
--- a/ams/core/matprocessor.py
+++ b/ams/core/matprocessor.py
@@ -311,8 +311,13 @@ class MatProcessor:
 
         Notes
         -----
-        Generator online status is NOT considered in its connectivity matrix.
-        The same applies for load, line, and shunt.
+        Generator online status is NOT considered in ``Cg`` — it is a
+        purely structural gen-to-bus map, because generator commitment
+        is a decision variable in the UC routines (status is applied via
+        ``ug``/``ugd`` instead). The load, line, and shunt connectivity
+        matrices (``Cl``, ``Cft``, ``Csh``) currently DO filter by their
+        element ``u`` status; reconciling them with this contract is a
+        tracked follow-up.
 
         Returns
         -------
--- a/ams/core/matprocessor.py
+++ b/ams/core/matprocessor.py
@@ -372,15 +372,19 @@ class MatProcessor:
         ng = system.StaticGen.n
 
         # bus indices: idx -> uid
+        # NOTE: generator online status is intentionally NOT applied here.
+        # Cg is a purely structural gen-to-bus map (see ``build`` Notes).
+        # Commitment/status is enforced elsewhere via ``ug``/``ugd`` (e.g.
+        # the ``pg`` bounds force an offline unit to zero output). Baking
+        # status into Cg would make a unit that is offline at build time
+        # permanently invisible to the network even after UC re-commits
+        # it via ``ugd`` — the root cause of the UC2ES/UCES divergence.
         idx_gen = system.StaticGen.get_all_idxes()
-        u_gen = system.StaticGen.get(src='u', attr='v', idx=idx_gen)
-        on_gen = np.flatnonzero(u_gen)  # uid of online generators
-        on_gen_idx = [idx_gen[i] for i in on_gen]  # idx of online generators
-        on_gen_bus = system.StaticGen.get(src='bus', attr='v', idx=on_gen_idx)
+        gen_bus = system.StaticGen.get(src='bus', attr='v', idx=idx_gen)
 
-        row = np.asarray(system.Bus.idx2uid(on_gen_bus))
-        col = np.asarray(system.StaticGen.idx2uid(on_gen_idx))
-        self.Cg._v = sps.csr_matrix((np.ones(len(on_gen_idx)), (row, col)), (nb, ng))
+        row = np.asarray(system.Bus.idx2uid(gen_bus))
+        col = np.asarray(system.StaticGen.idx2uid(idx_gen))
+        self.Cg._v = sps.csr_matrix((np.ones(len(idx_gen)), (row, col)), (nb, ng))
         self.Cg.col_names = idx_gen
         self.Cg.row_names = system.Bus.idx.v
         return self.Cg._v
--- a/ams/routines/uc.py
+++ b/ams/routines/uc.py
@@ -7,7 +7,8 @@ import numpy as np
 import pandas as pd
 
 from ams.core.param import RParam
-from ams.core.service import (NumOp, NumOpDual, MinDur, MinDurInit)
+from ams.core.service import (NumOp, NumOpDual,
+                              MinDurWindow, MinDurInit)
 from ams.utils.func import multiply_left_t
 from ams.routines.dcopf import DCOPF
 from ams.routines.rted import RTEDBase
@@ -232,19 +233,24 @@ class UC(SRBase, NSRBase, MPBase, RTEDBase, DCOPF):
                                  e_str='zug - Mzug * ugd <= 0')
 
         # --- minimum ON/OFF duration ---
-        # Interior window (Phase 3 of uc_min_on_off will replace
-        # MinDur with a Rajan-Takriti window-sum builder; today the
-        # don/doff cells reduce to v <= u, already implied by S1).
-        self.Con = MinDur(u=self.pg, u2=self.td1,
-                          name='Con', tex_name=r'T_{on}',
-                          info='minimum ON coefficient',)
+        # Rajan-Takriti window-sum form (interior, t >= TU_g - 1):
+        #   min-up : Σ_{s ∈ [t-TU+1, t]} v[g, s] ≤ u[g, t]
+        #   min-dn : Σ_{s ∈ [t-TD+1, t]} w[g, s] ≤ 1 - u[g, t]
+        # `Wup` / `Wdn` are sparse (n_g*n_t, n_g*n_t) block-diagonal
+        # matrices; `cp.vec` flattens vgd/wgd, the matmul produces
+        # window sums, and `cp.reshape` returns to (n_g, n_t).
+        self.Wup = MinDurWindow(u=self.pg, u2=self.td1,
+                                name='Wup', tex_name=r'W_{up}',
+                                info='min-ON window-sum coefficient',)
         self.don = Constraint(info='minimum online duration',
-                              name='don', e_str='cp.multiply(Con, vgd) - ugd <= 0')
-        self.Coff = MinDur(u=self.pg, u2=self.td2,
-                           name='Coff', tex_name=r'T_{off}',
-                           info='minimum OFF coefficient',)
+                              name='don',
+                              e_str='cp.reshape(Wup @ cp.vec(vgd, order="C"), vgd.shape, order="C") - ugd <= 0')
+        self.Wdn = MinDurWindow(u=self.pg, u2=self.td2,
+                                name='Wdn', tex_name=r'W_{dn}',
+                                info='min-OFF window-sum coefficient',)
         self.doff = Constraint(info='minimum offline duration',
-                               name='doff', e_str='cp.multiply(Coff, wgd) - (1 - ugd) <= 0')
+                               name='doff',
+                               e_str='cp.reshape(Wdn @ cp.vec(wgd, order="C"), wgd.shape, order="C") - (1 - ugd) <= 0')
 
         # Initial-state min-up/down: lock leading periods based on
         # how long each unit has already been in its current state
--- a/ams/routines/uc.py
+++ b/ams/routines/uc.py
@@ -256,7 +256,8 @@ class UC(SRBase, NSRBase, MPBase, RTEDBase, DCOPF):
                                 info='min-OFF window-sum coefficient',)
         self.doff = Constraint(info='minimum offline duration',
                                name='doff',
-                               e_str='cp.reshape(Wdn @ cp.vec(wgd, order="C"), wgd.shape, order="C") - (1 - ugd) <= 0')
+                               e_str=('cp.reshape(Wdn @ cp.vec(wgd, order="C"), wgd.shape, '
+                                      'order="C") - (1 - ugd) <= 0'))
 
         # Initial-state min-up/down: lock leading periods based on
         # how long each unit has already been in its current state
--- a/ams/routines/uc.py
+++ b/ams/routines/uc.py
@@ -7,7 +7,7 @@ import numpy as np
 import pandas as pd
 
 from ams.core.param import RParam
-from ams.core.service import (NumOp, NumOpDual, MinDur)
+from ams.core.service import (NumOp, NumOpDual, MinDur, MinDurInit)
 from ams.utils.func import multiply_left_t
 from ams.routines.dcopf import DCOPF
 from ams.routines.rted import RTEDBase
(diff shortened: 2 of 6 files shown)


ams/routines/uc.py
```diff
@@ -163,29 +172,36 @@ def __init__(self, system, config, **kwargs):
...
-                                e_str='-ugd[:, 0] + ug[:, 0] - wgd[:, 0] == 0',)
+        # State equation `u[t] - u[t-1] = v[t] - w[t]` (Rajan-Takriti
+        # 3-bin coupling, S1) plus initial-period anchor (S0) and
+        # exclusivity (X). Replaces the prior pair of signed
+        # equalities, which under boolean v/w forced ugd constant.
+        self.state = Constraint(name='state', info='commit state equation',
+                                e_str='ugd @ Mr - vgd[:, 1:] + wgd[:, 1:] == 0',)
+        self.state0 = Constraint(name='state0', info='initial commit state equation',
+                                 e_str='ugd[:, 0] - ug[:, 0] - vgd[:, 0] + wgd[:, 0] == 0',)
+        self.vwexcl = Constraint(name='vwexcl', info='startup/shutdown exclusivity',
```
leocuisine, line 209: **Finding 1 (correctness-risk) — relevant to this region but the actual line is uc.py:304 (objective), which is outside the diff so I anchor it here.**

The no-load cost term `cost += '+ cp.sum(cp.multiply(ug, c0) @ tlv)'` (uc.py:304) charges `c0` against the *frozen* initial `ug`, not the now-time-varying `ugd` that this PR unfreezes. A unit the optimizer commits where `ug0 == 0` pays no no-load cost; one it shuts down where `ug0 == 1` keeps paying. Harmless only because the demo case has `c0 == 0`, so no test catches it. Consider `cp.multiply(ugd, c0)`.

**Finding 2 (correctness-risk/docs) — this NOTE block.** The rationale that nonneg `csu`/`csd` make `v = max(Δu,0)` only holds for *strictly positive* costs; the demo case has `csu == csd == 0`, so during constant-commitment periods `vgd`/`wgd` can be fractional/non-unique. Not a hard bug (the state eq pins transitions and the window-sum only tightens), but the stated justification is inaccurate for the validated case.

ams/core/matprocessor.py
```diff
@@ -372,15 +372,19 @@ def build_cg(self):
         ng = system.StaticGen.n
 
         # bus indices: idx -> uid
+        # NOTE: generator online status is intentionally NOT applied here.
+        # Cg is a purely structural gen-to-bus map (see ``build`` Notes).
+        # Commitment/status is enforced elsewhere via ``ug``/``ugd`` (e.g.
+        # the ``pg`` bounds force an offline unit to zero output). Baking
```
leocuisine, line 383: **Finding 3 (docs).** This NOTE points readers to the `build()` Notes as authoritative, but that Notes block (matprocessor.py:314-315) claims status is ignored for `Cg` **and** load/line/shunt — yet `build_cl`/`build_csh`/`build_cft` still filter by element `u`. After this PR the contract is true only for `Cg`. Please scope the `build()` Notes to `Cg`, or add a 'currently only Cg' caveat, so the docstring isn't misleading as merged.

leocuisine (commented):
## Independent review — PR #268 (UC min on/off duration + structural `Cg`)

I read the full diff plus the surrounding source/tests, re-derived the Rajan–Takriti
formulation, and ran the new + modified test modules locally
(`tests/routines/test_uc_min_on_off.py` + `test_scenarios_milp_multiperiod.py`:
**54 passed**; UCES == UC2ES == 368.2068 reproduced). The core mechanics are sound:
the 3-binary state equation (`ugd @ Mr - vgd[:,1:] + wgd[:,1:] == 0`) is the textbook
coupling, the window-sum matrix indexing matches `cp.vec(..., order="C")`, and the
structural-`Cg` change is safe because UC/DCOPF/ED all pin an offline unit's `pg` to
zero through the `ugd`/`ctrle`-scaled capacity bounds (verified: offline gen ⇒
`pmaxe = pmine = 0`). Nice catch on the root cause of the UCES/UC2ES mismatch.

That said, unfreezing commitment exposed two modeling issues the PR does not address,
plus some doc/justification inaccuracies. Findings below, anchored and classified.

---

### 1. `correctness-risk` — no-load cost `c0` is charged against frozen `ug`, not time-varying `ugd`
`ams/routines/uc.py:304` — `cost += '+ cp.sum(cp.multiply(ug, c0) @ tlv)'`.
`ug` is the *constant* initial `StaticGen.u` broadcast across the horizon via `tlv`,
not the now-time-varying decision `ugd`. Before this PR commitment was frozen
(`ugd ≈ ug`), so this was harmless. Now that `ugd` genuinely varies, the fixed/no-load
cost is wrong in both directions: a unit the optimizer brings online at periods where
`ug0 == 0` pays *no* no-load cost, and a unit it shuts down where `ug0 == 1` keeps
paying. This directly undermines the "commitment can vary" premise for fixed-cost
accounting and silently shifts the optimal commitment for any case with nonzero `c0`.
It is latent only because the demo case has `c0 = 0` (verified) — so no test catches it.
Recommend `cp.multiply(ugd, c0)` (or document explicitly why `c0` tracks initial status).

### 2. `correctness-risk` / docs — `vgd`/`wgd` continuous-relaxation rationale doesn't hold when `csu = csd = 0`
`ams/routines/uc.py:176-181` (NOTE block). The justification — "nonneg costs `csu, csd`
in the objective … the optimum has `v = max(Δu,0)`, `w = max(-Δu,0)` automatically" —
only holds when `csu`/`csd` are **strictly positive**. The demo case the PR validates
against has `csu = csd = 0` (verified). With zero costs, during constant-commitment
periods (`u[t] == u[t-1]`) the state eq only forces `v[t] == w[t]`, and exclusivity
`v + w <= 1` leaves any `v = w ∈ [0, 0.5]` feasible — so `vgd`/`wgd` can be reported as
fractional/non-unique. This does **not** break min-up/down enforcement (the window-sum
`Σv ≤ u` only tightens with extra `v`, and transitions still pin `v,w` integral), so
it's not a hard bug — but the stated rationale is inaccurate for the validated case, and
downstream consumers reading `vgd`/`wgd` may see non-integral startup/shutdown actions.
Either keep them boolean, or soften the NOTE to "safe regardless of cost because the
state eq pins transitions; interior values are non-unique when start/stop costs are zero."

### 3. `docs` — `build()` Notes now over-promises for `Cl`/`Csh`/`Cft`
`ams/core/matprocessor.py:314-315`: "Generator online status is NOT considered in its
connectivity matrix. The same applies for load, line, and shunt." After this PR `build_cg`
honors that contract, but `build_cl` (`:409-416`), `build_csh`, and `build_cft` **still
filter by element `u`** (e.g. `on_load = np.flatnonzero(u_load)`). The new NOTE in
`build_cg` (`:375`) even points readers to "build Notes" as authoritative. So the
contract is now stated as true for all four but is actually only true for `Cg`. The PR
acknowledges this in its "follow-ups," but the in-tree docstring is misleading as merged —
please scope the Notes to `Cg` (or add a "currently only `Cg`; see #<follow-up>" caveat).

### 4. `test-gap` — rewritten `test_trip_gen` invariant is near-tautological and no longer tests a forced gen outage
`tests/routines/test_scenarios_milp_multiperiod.py:102-125`. The new invariant
`ugd < 0.5 ⇒ pg == 0` is essentially guaranteed by construction: `pg`'s bounds are
`pmine/pmaxe = (... ) * ugd` (`uc.py:161-168`), so `ugd == 0` forces `pg == 0` for *any*
feasible point — the test can only fail if the solver violates its own bounds. I confirmed
the solved solutions do have off-cells (70/62/62), so it isn't vacuous, but it exercises a
bound the solver cannot break rather than a behavior. More importantly, the suite no longer
has *any* test of a genuine forced generator outage — the old test (set `StaticGen.u = 0`,
assert `pg == 0`) was the only one. I accept the rationale that UC re-commits a guessed-off
unit, but consider also asserting something the optimizer *chooses* (e.g. that
total committed capacity still covers load, or that at least one unit the warm-start turned
off is legitimately re-committed in UCES/UC2ES) so the test has teeth beyond a bound check.
The removed UC2ES xfails are properly replaced by full obj/ugd/pg/aBus/plf asserts in
`test_align` — that part is solid.

### 5. `nit` — `MinDur` labeled "deprecated" in the index but not in its own docstring / no `DeprecationWarning`
`ams/core/service.py:24` lists `MinDur` as "(deprecated; see note)", but the class
docstring (`:920`) and `__init__` carry no deprecation note or `DeprecationWarning`, and
nothing in the diff schedules its removal. Either wire up the deprecation (warning +
"removal in vX") to match the sibling services that were deprecated this cycle, or drop the
"deprecated" label until that's done.

### 6. `nit` — `MinDurWindow.v` builds the band matrix with a triple Python loop
`ams/core/service.py:1065-1072`. `for g → for t → for s` is `O(n_g · n_t · TU)` pure-Python
appends, re-run on every `update()` that re-finalizes (any `no_parse=True` param change
forces re-evaluate, `routine.py:1248-1258`). Fine at `5 gens × 24 × TU`, but AMS leans on
sparse/vectorized matrix builds elsewhere; a vectorized diagonal-band construction would
scale better for larger UC horizons. Optional.

---

Items 1 and 3 are the ones I'd want resolved (or explicitly deferred with a tracking
issue) before merge; 1 is a real objective bug masked only by `c0 == 0` on the demo case.
Everything else is advisory. Inline comments left for findings 1 and 3.

ams/routines/uc.py
leocuisine: Good catch — this is a real objective bug, fixed in 10e89e7b. The no-load term now charges `cp.sum(cp.multiply(c0, ugd))` (per committed period) instead of `cp.multiply(ug, c0) @ tlv` against the frozen initial `ug`. It was incidentally correct before Phase 1 (when `ug` *was* the commitment) and latent because the demo case has c0==0. Verified DPP-clean and that the no-load cost now responds to commitment with a nonzero c0.

ams/core/matprocessor.py
leocuisine: Fixed in 10e89e7b — the `build()` Notes now state that only `Cg` is structural (status-independent) and that `Cl`/`Cft`/`Csh` still filter by element status, flagged as a follow-up. The generator case is special because commitment is a decision variable (`ugd`); load/line/shunt status are not decisions in the current routines, so their filtering is self-consistent — reconciling the full contract is tracked separately.

leocuisine:
Thanks for the thorough review. Triage of all six findings:

**1 (correctness-risk) — no-load cost `c0` charged against frozen `ug`:** ✅ Fixed in 10e89e7b. Real objective bug, latent only because the demo case has `c0==0`. Now `cp.sum(cp.multiply(c0, ugd))`.

**2 (correctness-risk/docs) — `vgd`/`wgd` relaxation rationale:** ✅ Comment corrected in 10e89e7b. You're right that the "nonneg costs force v=max(Δu,0)" claim only holds for strictly-positive `csu`/`csd`. The relaxation is still valid: at transitions the state equation + exclusivity pin (v,w) to {0,1} structurally; on steady periods zero costs leave a [0,0.5] slack, but that slack is free w.r.t. the binary `ugd` (the min-dur window sums stay satisfiable by the canonical v=max(Δu,0)), so the optimal commitment is unaffected. Comment now says this.

**3 (docs) — `build()` Notes mismatch:** ✅ Fixed in 10e89e7b — Notes now scope the status-independence claim to `Cg` only and flag `Cl`/`Cft`/`Csh` as a tracked follow-up.

**4 (test-gap) — `test_trip_gen` near-tautological, no genuine outage test:** Acknowledged, keeping as-is. UC *optimizes* commitment — there is no forced-outage input (no per-slot status; `StaticGen.u` only seeds `ug0`; `pmax=0` makes the small case infeasible via reserves), so a genuine-trip test isn't applicable the way it is for the LP/PF families. The `ugd=0 ⇒ pg=0` invariant is intentionally the portable check; the real regression for this PR's behavior change is the restored full `test_align[UC2ES]` (obj/ugd/pg/plf), which the removed xfails are replaced by.

**5 (nit) — `MinDur` labeled deprecated without a warning:** Acknowledged. Full `DeprecationWarning` + removal is deferred to the v1.4.0 service-module cleanup (consistent with how the other UC services are being phased).

**6 (nit) — `MinDurWindow.v` triple Python loop:** Acknowledged. Correctness-first here; vectorizing the band-matrix build is a worthwhile perf follow-up but the matrix is built once per init, not per solve, so it's not on the hot path.

A separate follow-up is also logged for the **vacuous PTDF line limits** (UC2/UC2ES limits reference `Bf@aBus` while the PTDF routines drop the nodal balance pinning `aBus`) — pre-existing, surfaced during this work, and out of scope here.

